PR #8, opened 2020-11-07
Todo #1

hategan:
Fixed some of the items in the TODO list. Specifically:

* Add metadata to JobStatus. One important use case we discussed is...
* Add a get_version method/function with a note about version obj vs string depending on programming  language
* add ability to have custom attributes which could be passed to the underlying LRM (aka. dynamic attributes).
* merge prev() spirit into previous method (the order)

I'm issuing this PR instead of waiting because it seems like the probability of conflicts is getting higher. In fact, we're already hitting a bit of a conflict with #1 

Anyway, please review, let's resolve the contentious parts, and merge the rest.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -588,7 +588,7 @@ not required to do so.
 - getters/setters can be replaced by properties, depending on what is
 customary in the language in which the library is implemented
 
-- the root of the class hierarchy is called `Object` in this document
+
 
 ### JobExecutor
 
--- a/specification.md
+++ b/specification.md
@@ -94,9 +94,6 @@ nicely, with a metadata dictionary.
 - [x] Add a get_version method/function with a note about version obj vs
 string depending on programming  language
 
-- [x] add ability to have custom attributes which could be passed to the
-underlying LRM (aka. dynamic attributes).
-
 - [x] merge prev() spirit into previous method (the order)
 
 - [ ] "canceled" or "cancelled"?
@@ -107,6 +104,9 @@ distinguish between EAGAIN types of errors and others.
 - [ ] think more about env var expansion in arguments and other places.
 The important issue is how much of a burden this is on implementations if
 we mandate it.
+
+- [ ] add ability to have custom attributes which could be passed to the
+underlying LRM (aka. dynamic attributes).
 
 - [ ] we need to go through the resource spec; many common things
 supported by other JM APIs are not supported by Flux Jobspec V1, such as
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -1294,6 +1315,19 @@ resources reserved through the advanced reservation represented by this
...
+Object? getCustomAttribute(String name);
+```
+
+Allows setting/querying of custom attributes. Implementations should use a
+dictionary to implement the custom attributes. Additionally, implementations are
+encouraged to make sensible decisions on whether to store some or all of the
+fixed attributes in the same dictionary or not. It is, therefore, entirely
+possible for `getCustomAttribute("duration")` to return a value passed earlier
+to `setDuration()`, although the specific custom attribute name need not be
+`"duration"`.
```
SteVwonder, line 1330: I'm a little bit concerned about violating the principle of least surprise if a user happens to use a key that they didn't realize was reserved/tied to a fixed attribute. It's admittedly a pretty low probability with the functions we have defined currently.  Thoughts on allowing `get`s of the fixed attributes but not `set`s via this API?

SteVwonder (commented):
Generally LGTM!  One comment below about the custom attribute interface.

specification.md
andre-merzky: I echo that concern actually.  I would think that we don't need to make these attributes writeable by the user of the API, but rather expect the implementation to provide read-only values.

andre-merzky: I would also suggest to return the custom attributes in one call.  Otherwise we would need to add inspection to inform the application about what attributes can be queries at that point (which may differ from one `JobStatus` object to the next).  So we could return a map, or a set of key/value pairs.  Well, I am a bit biased I guess -- see the version in #1 which addresses the same TODO a bit differently.

andre-merzky: Ignore my previous (deleted) comments for this method, I missed the point.  Let me instead please ask what the use case this method is trying to address?

hategan: The assumption here is that users will not randomly set keys in this dictionary for no reason. And a user would have no reason to use a key unless an implementation assigned some function to that key.

hategan: > Ignore my previous (deleted) comments for this method, I missed the point. Let me instead please ask what the use case this method is trying to address?

@andre-merzky I can't remember where this came from. I can't find it in the comments. @SteVwonder any ideas?

SteVwonder: I don't see anything in the google docs notes either.  I remember us mentioning that it is easy for people to abuse this functionality, but I can't remember what the specific intended use case was.

andre-merzky: I would not mind removing that method then until we are clear about the use case.

hategan: Removed

hategan:
Can I merge this?

SteVwonder (approved):
LGTM!

PR #25, opened 2020-11-07
State model

andre-merzky:
- add state model description when introducing the `Job` class
  - rename `terminal state` to `final state` (this is a well defined term AFAICT)
  - remove `pred()` as discussed
  - clarify `isGreaterThan`


specification.md
```diff
@@ -694,6 +695,53 @@ to `null`.
...
 ### Job
 
+#### State Model
+
+Job instances are, in this API, stateful objects.  A job's state can be
+inspected via the `job.getStatus()` method which will return a `JobStatus`
+instance on which the job's state is available as an attribute.  State
+transitions can also be received via callbacks
+
+TODO: how are callbacks registered?
```
SteVwonder, line 705: I believe the `setJobStatusCallback` method of the `Executor` or the `setStatusCallback` of the `Job`.

specification.md
```diff
@@ -694,6 +695,53 @@ to `null`.
...
+execution with an error code, but can also indicate a backend error, or
+a library error of any kind.  The `FAILED` state is final.
+
+At any point in time (until the job is final), the job can enter the `CANCELED`
+state as reaction to the `job.cancel()` call.  Note that the transition to
+`CANCELED` is not immediate when calling that method, but the state transition
+only occurs once the backend is enacting that request.  
+
+The `ACTIVE` state is the only state where the job will consume resources.
+
```
SteVwonder, line 725: I don't see any definition for PENDING or RESUMED but  they are in the state model diagram.  Not sure which one is right (the diagram or the text).

specification.md
```diff
@@ -957,21 +1005,20 @@ String? getMessage()
...
+[`status.getState().isFinal()`](#jobstate-isFinal).
 
 
 
 ### JobState
 
 An enumeration holding the possible job states, which are: `NEW`,
-`QUEUED`, `ACTIVE`, `SUSPENDED`, `RESUMED`, `COMPLETED`, `FAILED`,
-`CANCELLED`.
+`QUEUED`, `ACTIVE`, `COMPLETED`, `FAILED`, and `CANCELLED`.
```
SteVwonder, line 1021: If we intentionally dropped `SUSPENDED` and `RESUMED`, they should be dropped from the state diagram too.

The state diagram also still has `DONE` rather than `COMPLETED`.

SteVwonder (commented):
Generally LGTM! Main comments are about inconsistencies between the text and the state diagram.

specification.md
andre-merzky: yes, of course (*blush*)

specification.md
andre-merzky: Yeah, that suspend/resume is inconsistent, will fix.

specification.md
andre-merzky: Ack

specification.md
andre-merzky: suspend/resume are gone.

specification.md
andre-merzky: done

SteVwonder:
Text looks good! I think the state model svg just needs to be updated and this is good to go IMO.

andre-merzky:
> Text looks good! I think the state model svg just needs to be updated and this is good to go IMO.

Ah, of course - that is updated now.
Thanks!

SteVwonder (approved):
Excellent! LGTM!

SteVwonder:
Since we went over this is detail on a Friday call, I think this is good to merge now. Thanks @andre-merzky!

Issue #23, opened 2020-11-07
JobExecutor constructor missing


andre-merzky:
I suggest to add a constructor which takes a backend URL as single argument, to identify the backend to connect to for job management.  The JobExecutor has a method `getName()` - I would suggest to change that to `getUrl()` to obtain that URL (or render that as an attribute on the object really):

```py
je = JobExecutor('slurm://localhost')
assert je.url == 'slurm://localhost'
```

hategan:
I do not understand the purpose of the URL, since there is only localhost in Layer 0.

I also do not understand your code above. JobExecutor is meant as an abstract base class which cannot be instantiated directly. It would have subclasses, such as SlurmJobExecutor. Getting a specific instance of an executor would be done through a factory and the name would be relevant there:

```python
je = JobExecutorFactory.getExecutor('slurm')
assert je.name == 'slurm'
```

In Layer 1, for remote execution, I would also assume that we would use a single instance of a remote executor (of a certain type) to route jobs to different machines, so the precise location would not be tied to the executor instance, but to the job that is being submitted.

andre-merzky:
> I do not understand the purpose of the URL, since there is only localhost in Layer 0.

An URL does not imply remote.  We can use some other format of endpoint specifier, like an unqualified string - but I assume *some* form of an endpoint specifier would be needed to let the implementation know where the backend is running (locally) it needs to connect to, and what type of backend it is.

> ... JobExecutor is meant as an abstract base class
> ... Getting a specific instance of an executor would be done through a factory

Hmm, I can see that this was clear in your mental image of the API, but I did not infer that from the text.  Why would a factory be different from a constructor?  That is really just an implementation detail again, isn't it?  A factory can be wrapped into an Constructor via delegation; a constructor can be called via a factory wrapper.  I can see that factories are probably the way to go in Java, but are not as frequently used in other languages.  Well, anyway, matter of taste, fine either way :-)

Either way though, one would need some argument which specifies not only backend type, but also the how to connect to it: be it a local port, or a socket in the file system, or whatever.  For backends like Slurm which run in system space, that knowledge may be implicitly available (a `qstat` command will connect to slurm w/o additional info), but a user space Flux instance will need the Flux-URL to connect to it's socket, a user space RCT component needs a ZMQ port to talk to, Fireworks needs a mongodb location (with access auth), etc - those endpoints are not obvious and not unique.

hategan:
[...]

PR #26, opened 2020-11-07
add JobExecutor constructor

andre-merzky:
As the title says: this adds a `JobExecutor` constructor, as per #23.


specification.md
```diff
@@ -595,6 +595,16 @@ customary in the language in which the library is implemented
 #### Methods
 
 
+<a name="jobexecutor-constructor"></a>
+```java
+JobExecutor(String url)
+```
+
+Instantiates an `JobExecutor` object and binds it to the backend specified by
+URL `url`.  The URL will also encode any authentication details needed to
```
SteVwonder, line 604: Could this constructor throw an exception if it fails to connect, either due to authentication or due to a non-responsive/missing/dead executor?

andre-merzky: Yes, that makes sense of course - will add.

hategan: I think we need to hit the breaks a bit since things seem to be getting a bit out of control. Remote connections are beyond the scope of what we are doing. If we are talking about them, something has gone wrong.

andre-merzky: See note in #23 - I don't think URL needs to mean remote.  Feel free to replace `URL` with `endpoint specifier` or whatever.

SteVwonder: Yeah, I interpreted this to be an implementation/executor-specific URL that doesn't necessarily need to be remote.  In the case of Flux, I expect it to be the [URI that you would normally pass to `flux_open`](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man3/flux_open.html?highlight=flux_open#description) (e.g., `file:///tmp/flux-ndacld8763/local/0`).  Useful for when you have a bunch of Flux instances running at once.  In Slurm's case, maybe it is the location of your `slurm.conf` (assuming you have it in a non-default location or multiple slurm instances running on your cluster for whatever reason).

Side benefit that in Flux's case (and I'm sure others), it could easily be adapted to the remote case (e.g., `ssh://stephen@mycluster/path/to/domain-socket`).

SteVwonder: PS - I just saw the discussion over in #23, and my comment above doesn't add much to what was already discussed there.  So don't worry about rebutting my comment here.

andre-merzky: Added the exception.

SteVwonder (approved):
This LGTM! But I know @hategan had a lot of reservations about the URL expressed in #23.  So I don't necessarily feel comfortable merging this with him looking at it first.

@hategan: would you like to make a pass on this before it is merged?

hategan:
Right. JobExecutor is meant as a base class/interface. One could also see it as a frontend that delegates to a specific instance of another class that implements an identical interface, but that is a contrived solution that goes against the straightforward OOP solution for this problem. In Python, one could duck-type around this. However, all of this is unreasonable to ask of all implementations.

We talked, I believe, about the ability to configure a job executor. That could be a place where we would allow some generic configuration dictionary to be passed to an executor and we may chose to have that done during construction. However, constructors are not generally inherited in OO languages. So the constructor of an ABC seems like the wrong place to specify something like that. Even if we do that, an URL seems like a rather limiting way of specifying a configuration, unless we use URL parameters, which... I mean we can ultimately encode anything in base64 encoded XML if we really want to...

I think we should open an issue about configuring executors. There are a number of options that we could discuss, such as entirely stateless executors, configurations passed to a factory, or a mix of the two.

Anyway, in short, I do not think that a URL passed to the constructor of an ABC is anywhere near the right way to go, except, perhaps, if writing a thin layer on top of an existing jpsi/saga implementation.

andre-merzky:
> Right. JobExecutor is meant as a base class/interface. One could also see it as a frontend that delegates to a specific instance of another class that implements an identical interface, but that is a contrived solution that goes against the straightforward OOP solution for this problem. In Python, one could duck-type around this. However, all of this is unreasonable to ask of all implementations.

I am in a somewhat constant state of confusion about those discussions.  I kind of understand that you intent the document to guide different aspects (API spec, language binding, implementation) - but the topics seem to always interfere and intermix.  

What the PR tried to provide is a mechanism to instantiate a defined class object, providing an endpoint address (URL or String) as input.   

> However, constructors are not generally inherited in OO languages

That is a language issue, and needs to be addressed in the language where that is the case.  Other languages do inherit constructors, and others don't have constructors, and others favour factories.  So what - that should not stop is from defining instantiation parameters.  Languages may have renderings as ABC (or not), implementations may use delegation (or not).  What has all that to do with defining an abstract API?

I don't know how duck-typing comes into the discussion on that level either. 

> I think we should open an issue about configuring executors. There are a number of options that we could discuss, such as entirely stateless executors, configurations passed to a factory, or a mix of the two.

Unless we *specify* that an executor is stateless, this is an implementation issue.  Also, the fact if a class is stateless or not has nothing to do with the question of defining a constructor?

> Anyway, in short, I do not think that a URL passed to the constructor of an ABC is anywhere near the right way to go, except, perhaps, if writing a thin layer on top of an existing jpsi/saga implementation.

Which would be a valid implementation I would think?

hategan:
> > Right. JobExecutor is meant as a base class/interface. One could also see it as a frontend that delegates to a specific instance of another class that implements an identical interface, but that is a contrived solution that goes against the straightforward OOP solution for this problem. In Python, one could duck-type around this. However, all of this is unreasonable to ask of all implementations.
> 
> I am in a somewhat constant state of confusion about those discussions. I kind of understand that you intent the document to guide different aspects (API spec, language binding, implementation) - but the topics seem to always interfere and intermix.

Surely an issue of miscommunication...

> 
> What the PR tried to provide is a mechanism to instantiate a defined class object, providing an endpoint address (URL or String) as input.

There is no endpoint in Layer 0 and we need to discuss how we specify endpoints and other issues (such as security info) in other layers, but I'm not sure why we're doing that here.

> 
> > However, constructors are not generally inherited in OO languages
> 
> That is a language issue, and needs to be addressed in the language where that is the case. 

So then why on earth would you have the specification mandate something that doesn't make sense in some of the languages?

> Other languages do inherit constructors, and others don't have constructors, and others favour factories. So what - that should not stop is from defining instantiation parameters. Languages may have renderings as ABC (or not), implementations may use delegation (or not). What has all that to do with defining an abstract API?

I don't know. You tell me why you're trying to push for one particular solution in the abstract API.

> 
> I don't know how duck-typing comes into the discussion on that level either.

Because of your suggestion that the JobExecutor not be a base class. The only situation in which you could have multiple implementations of the same API be readily usable without them formally implementing the same type is when duck typing applies.

> 
> > I think we should open an issue about configuring executors. There are a number of options that we could discuss, such as entirely stateless executors, configurations passed to a factory, or a mix of the two.
> 
> Unless we _specify_ that an executor is stateless, this is an implementation issue. Also, the fact if a class is stateless or not has nothing to do with the question of defining a constructor?

It has to do with specifying settings for an instance. For example, we could have an SSH executor instance that is bound to a specific endpoint at construction, or we could have an executor that is not and takes the location information from somewhere else, such as the job specification.

> 
> > Anyway, in short, I do not think that a URL passed to the constructor of an ABC is anywhere near the right way to go, except, perhaps, if writing a thin layer on top of an existing jpsi/saga implementation.
> 
> Which would be a valid implementation I would think?

Perhaps so in Python and as a thin layer on top of an existing implementation, but not as a language agnostic specification for all kinds of implementations.

andre-merzky:
> Surely an issue of miscommunication...

Yes :-)
 
> > What the PR tried to provide is a mechanism to instantiate a defined class object, providing an endpoint address (URL or String) as input.
> 
> There is no endpoint in Layer 0 and we need to discuss how we specify endpoints and other issues (such as security info) in other layers, but I'm not sure why we're doing that here.

I am not a maintainer of any LRM, but want to implement this API to abstract SLURM and FLUX.  The PR proposes that some form of parameter is passed on construction to tell the implementation what LRM to talk to, how to contact the specific FLUX instance, etc.  Not having that ability would make it impossible for me to implement this as a replacement for SAGA, which is where we started.
 
> > > However, constructors are not generally inherited in OO languages
> > That is a language issue, and needs to be addressed in the language where that is the case.
> 
> So then why on earth would you have the specification mandate something that doesn't make sense in some of the languages?

I don't care about the mechanism used in this doc.  This doc uses a class, so I added a constructor for initializing it.  If it would have defined a function-based API, I would have added an `jpsi_init(url)` function, etc.  We can specify a factory - but that adds another object.  But sure, don't care.  I care about an initialization parameter.

 
> > Other languages do inherit constructors, and others don't have constructors, and others favour factories. So what - that should not stop is from defining instantiation parameters. Languages may have renderings as ABC (or not), implementations may use delegation (or not). What has all that to do with defining an abstract API?
> 
> I don't know. You tell me why you're trying to push for one particular solution in the abstract API.

see above.

 
> > I don't know how duck-typing comes into the discussion on that level either.
> 
> Because of your suggestion that the JobExecutor not be a base class. The only situation in which you could have multiple implementations of the same API be readily usable without them formally implementing the same type is when duck typing applies.

I don't think that is the case, or I am missing the point.  I can well have a multi-bound implementation w/o ABC and duck typing - see radical.saga, and I can easily switch implementations w/o duck typing - see MPI.

hategan:
> > Surely an issue of miscommunication...
> 
> Yes :-)
> 
> > > What the PR tried to provide is a mechanism to instantiate a defined class object, providing an endpoint address (URL or String) as input.
> > 
> > 
> > There is no endpoint in Layer 0 and we need to discuss how we specify endpoints and other issues (such as security info) in other layers, but I'm not sure why we're doing that here.
> 
> I am not a maintainer of any LRM, but want to implement this API to abstract SLURM and FLUX. The PR proposes that some form of parameter is passed on construction to tell the implementation what LRM to talk to, how to contact the specific FLUX instance, etc. Not having that ability would make it impossible for me to implement this as a replacement for SAGA, which is where we started.

`JobExecutor` is an abstract base class. You have adapters, `SLURMJobExecutor`, `FluxJobExecutor`, etc., which have `JobExecutor` somewhere in their inheritance tree.

I am, however, not sure what you mean by "a replacement for SAGA". It sounds like the replacement for SAGA is almost exactly SAGA?

> 
> > > > However, constructors are not generally inherited in OO languages
> > > > That is a language issue, and needs to be addressed in the language where that is the case.
> > 
> > 
> > So then why on earth would you have the specification mandate something that doesn't make sense in some of the languages?
> 
> I don't care about the mechanism used in this doc. This doc uses a class, so I added a constructor for initializing it. 

An interface/ABC.
Then we had a long discussion about delegation vs. ABCs. And now we're back to this discussion which seems to entirely ignore that discussion.

> If it would have defined a function-based API, I would have added an `jpsi_init(url)` function, etc. We can specify a factory - but that adds another object. 

And then we said we maybe shouldn't specify how an implementation should instantiate these concrete classes. But, this did come out in the other discussion, you cannot avoid a factory. You may be able to avoid it in the interface, but if you want to go from a string to an instance, you need something like a factory.

> But sure, don't care. I care about an initialization parameter.

Well, not if its sole purpose is to specify an implementation of JobExecutor. And then we discussed endpoints and settings, so then there's that separate discussion which we should have. No?

> 
> > > Other languages do inherit constructors, and others don't have constructors, and others favour factories. So what - that should not stop is from defining instantiation parameters. Languages may have renderings as ABC (or not), implementations may use delegation (or not). What has all that to do with defining an abstract API?
> > 
> > 
> > I don't know. You tell me why you're trying to push for one particular solution in the abstract API.
> 
> see above.
> 
> > > I don't know how duck-typing comes into the discussion on that level either.
> > 
> > 
> > Because of your suggestion that the JobExecutor not be a base class. The only situation in which you could have multiple implementations of the same API be readily usable without them formally implementing the same type is when duck typing applies.
> 
> I don't think that is the case, or I am missing the point. I can well have a multi-bound implementation w/o ABC and duck typing - see radical.saga, and I can easily switch implementations w/o duck typing - see MPI.

You have to re-compile if you want to switch MPI implementations. I hope that's not what we're advocating here. We need the ability to publish workflow software that uses JPSI in a way that does not require re-compilation every time clusters are changed.

I don't get it. You have a cpi.Service class in SAGA, which is pretty much an ABC, since it has empty methods and all specific services inherit from. JobExecutor is the equivalent of cpi.Service. You seem to be wanting to use JobExecutor as job.Service. I mean even SAGA is clear where the API is (CPI) and that job.Service is not in the CPI.

andre-merzky:
Yeah, we are running circles...

> I don't get it. You have a cpi.Service class in SAGA, which is pretty much an ABC, since it has empty methods and all specific services inherit from. JobExecutor is the equivalent of cpi.Service. You seem to be wanting to use JobExecutor as job.Service. I mean even SAGA is clear where the API is (CPI) and that job.Service is not in the CPI.

The SAGA CPI is absolutely an implementation detail, and is *not* the SAGA API.  As such, I don't know how that plays into the discussion.   And even here, the SAGA CPI base class (the abstract base for other abstract bases for adaptor implementations) has a [constructor](https://github.com/radical-cybertools/radical.saga/blob/devel/src/radical/saga/adaptors/cpi/base.py#L24), so I doubly don't know how that plays a role in this discussion.  I remain confused...

hategan:
> Yeah, we are running circles...
> 
> > I don't get it. You have a cpi.Service class in SAGA, which is pretty much an ABC, since it has empty methods and all specific services inherit from. JobExecutor is the equivalent of cpi.Service. You seem to be wanting to use JobExecutor as job.Service. I mean even SAGA is clear where the API is (CPI) and that job.Service is not in the CPI.
> 
> The SAGA CPI is absolutely an implementation detail, and is _not_ the SAGA API. As such, I don't know how that plays into the discussion. And even here, the SAGA CPI base class (the abstract base for other abstract bases for adaptor implementations) has a [constructor](https://github.com/radical-cybertools/radical.saga/blob/devel/src/radical/saga/adaptors/cpi/base.py#L24), so I doubly don't know how that plays a role in this discussion. I remain confused...

Yeah, I'm not sure what's happening here:
- your comment is ignoring the other issues in the discussion and focuses on only this one issue; do you agree or disagree with the other points? do they address your questions where applicable?
- all classes have constructors (although sometimes they are implicit); so I am not sure what purpose the statement that cpi.Service has a constructor is meant to serve; we are discussing whether a specific constructor should be part of the public API or not, not whether the classes in the API have constructors
- what implementation purpose does the cpi.Service class serve given that it only contains method signatures? the point here is that, even though cpi.Service is not user-facing, the pattern of adapters inheriting from a base class is also present in SAGA.
- the fact that SAGA chooses to use a single user-facing class for all adapters and *delegate* to subclasses of cpi.Service is something that should not be forced on everybody who is implementing the JPSI API

andre-merzky:
> Yeah, I'm not sure what's happening here:
> 
>     * your comment is ignoring the other issues in the discussion and focuses on only this one issue; do you agree or disagree with the other points? do they address your questions where applicable?

Just a tad tired, Mihael.  I can reply to the other parts - but as said, we do seem to run circles here, and I am not sure I am able to disentangle.  So I closed the PR and we'll see how that goes w/o it, ok?


>     * all classes have constructors (although sometimes they are implicit); so I am not sure what purpose the statement that cpi.Service has a constructor is meant to serve; we are discussing whether a specific constructor should be part of the public API or not, not whether the classes in the API have constructors

This is a statement that I, again, he difficulties to grasp.  Let me try to rephrase, to see if I got that right:  `all classes have constructors, but here we discuss if our classes have a constructor`.  I miss the logic in that.  If all classes have constructors, then our's do as well, or so I would think?  So maybe you mean if we wan to describe the constructor, or leave it open (and thus to implementers)?  But then the API will not lead to portable code, because, well, c'tor sigatures may differ.  


>     * what implementation purpose does the cpi.Service class serve given that it only contains method signatures? the point here is that, even though cpi.Service is not user-facing, the pattern of adapters inheriting from a base class is also present in SAGA.

Yeah, adaptors can inherit from a base class.  What has that to do with our API?  An adaptor is not a user facing software component (generally at least).  The SAGA CPI is not user facing.  Classes can inherit from other classes.  That also holds for user facing classes.  They can also inherit from ABC. 

Those are all truisms - what has that to do with the PR?  That PR suggests to specify an URL to initialize our Executor class in some way.  I really don't care about inheritance *at all* with this PR.  This is *just* about initializing the class instance.  


>     * the fact that SAGA chooses to use a single user-facing class for all adapters and _delegate_ to subclasses of cpi.Service is something that should not be forced on everybody who is implementing the JPSI API

Again, the SAGA CPI is not user facing, it is not part of the public API.  Even if it were though, how does that relate to the PR?  I don't want to force delegation nor ABC nor adaptors nor inheritance or any other implementation choice - happy to show a proof of concept implementation w/o any (or all) of those.  

Maybe let me take a different approach: assume we define a `File` class which is semantically supposed to represent a physical file.  I would assume that it might be useful if that class gets a constructor (or initializer or factory method) which accepts a file location (path) as mandatory parameter - because otherwise it would be kind of difficult to instantiate that class as a representation of a specific file.  Are you with me so far?  If so, would you include such a constructor (or initializer etc) in the `File` API specification?  If yes, what is the difference to our `psi.j.executor` which represents a specific existing LRM?  The PR proposes to add just such an constructor / initializer / factory to point the class instance to the specific LRM instance it is supposed to represent.

hategan:
> > Yeah, I'm not sure what's happening here:
> > 
> > * your comment is ignoring the other issues in the discussion and focuses on only this one issue; do you agree or disagree with the other points? do they address your questions where applicable?
> 
> Just a tad tired, Mihael. I can reply to the other parts - but as said, we do seem to run circles here, and I am not sure I am able to disentangle. So I closed the PR and we'll see how that goes w/o it, ok?

It can be tiresome, I agree. It would be much better if we could have a quick conversation in person. But that's not an option.
I think we should try to calmly figure this out because, even if we drop this PR, we're still going to have to address some of these issues. But if you want me to drop this conversation (temporarily or permanently), please let me know.

> 
> > * all classes have constructors (although sometimes they are implicit); so I am not sure what purpose the statement that cpi.Service has a constructor is meant to serve; we are discussing whether a specific constructor should be part of the public API or not, not whether the classes in the API have constructors
> 
> This is a statement that I, again, he difficulties to grasp. Let me try to rephrase, to see if I got that right: `all classes have constructors, but here we discuss if our classes have a constructor`. I miss the logic in that. If all classes have constructors, then our's do as well, or so I would think? So maybe you mean if we wan to describe the constructor, or leave it open (and thus to implementers)? But then the API will not lead to portable code, because, well, c'tor sigatures may differ.

The specification specifies an API, which consists of classes with methods and properties. The understanding is that implementing libraries will have those classes and those methods and those properties. The understanding IS NOT that implementing libraries will ONLY have those classes and methods and properties. Again, we are deciding what goes into the PUBLIC API, not what goes into the implementing classes.

So even though JobExecutor, as an ABC, may have a constructor, this does not imply that it needs to be specified by JPSI.

To the issue of portability: 
1. I think we still haven't decided the extend of source-level compatibility we want to have
2. we can either leave the mechanism for instantiation open to library implementers, or we can specify one
3. there are a number of options, such as factories, what SAGA is doing, or manual construction. 
You picked one of these solutions in this PR, implicitly answering both (2) and (3). I think we need to first discuss and come to an agreement on (2) and (3), because what you seem to have picked does not match what I or others I've talked to believe is the most common solution in this scenario. It is also fine to not have complete source-level compatibility as far as I'm concerned.

> 
> > * what implementation purpose does the cpi.Service class serve given that it only contains method signatures? the point here is that, even though cpi.Service is not user-facing, the pattern of adapters inheriting from a base class is also present in SAGA.
> 
> Yeah, adaptors can inherit from a base class.

Not "can", but "must" in strictly typed languages (notwithstanding contrived voodoo meant to bypass a mechanism for which a perfectly reasonable solution is already provided).

> What has that to do with our API? An adaptor is not a user facing software component (generally at least). The SAGA CPI is not user facing. Classes can inherit from other classes. That also holds for user facing classes. They can also inherit from ABC.

I was saying that JobExecutor is meant as an ABC and does not need an explicit constructor. You were saying that an ABC is not necessarily needed for executor adapters. I don't see how you can reasonably do what we want without one. And I don't see why the ABC would not be the public interface if we do have an ABC for executors.

> 
> Those are all truisms - what has that to do with the PR? That PR suggests to specify an URL to initialize our Executor class in some way. I really don't care about inheritance _at all_ with this PR. This is _just_ about initializing the class instance.

If JobExecutor is an ABC, it makes no sense to have a constructor for it since it can never be called directly by user code. So, in order to figure out whether a constructor for JobExecutor makes sens, we need to figure out whether JobExecutor is an ABC or not.

> 
> > * the fact that SAGA chooses to use a single user-facing class for all adapters and _delegate_ to subclasses of cpi.Service is something that should not be forced on everybody who is implementing the JPSI API
> 
> Again, the SAGA CPI is not user facing, it is not part of the public API. Even if it were though, how does that relate to the PR? I don't want to force delegation nor ABC nor adaptors nor inheritance or any other implementation choice - happy to show a proof of concept implementation w/o any (or all) of those.

You are forcing JobExecutor to not being an ABC if you define a public constructor for it.

> 
> Maybe let me take a different approach: assume we define a `File` class which is semantically supposed to represent a physical file. I would assume that it might be useful if that class gets a constructor (or initializer or factory method) which accepts a file location (path) as mandatory parameter - because otherwise it would be kind of difficult to instantiate that class as a representation of a specific file. Are you with me so far? If so, would you include such a constructor (or initializer etc) in the `File` API specification? If yes, what is the difference to our `psi.j.executor` which represents a specific existing LRM? The PR proposes to add just such an constructor / initializer / factory to point the class instance to the specific LRM instance it is supposed to represent.

Are you saying that `File("/etc/rc.local")` and `File("/bin/ls")` is somewhat equivalent to `JobExecutor("slurm://")` and `JobExecutor("torque://")`?

andre-merzky:
> Are you saying that `File("/etc/rc.local")` and `File("/bin/ls")` is somewhat equivalent to `JobExecutor("slurm://")` and `JobExecutor("torque://")`?

Yes, absolutely.   Take an example: I run two flux instances in userspace, and the same application wants to submit tasks to instance one or instance two.  That is a use case we implement, FWIW (multiple PRTE DVMs, multiple ORTE DMVs, multiple Flux instances, etc).  Now the application uses our API:
```py
je = psi.j.Executor()
je.submit(...)
```

there is no means to tell the implementation what Flux instance to talk to.  For a generic implementation (which can submit to Flux or PRRTE), I can't even let the implementation know that I intent to communicate with Flux.

You mentioned several times that layer 0 is kind of implicitly knowing where to connect to - and I replied multiple times that this is not working for us.  So lets skip that loop this time :-)

> If JobExecutor is an ABC, it makes no sense to have a public constructor for it since it can never be called directly by user code. So, in order to figure out whether a constructor for JobExecutor makes sense, we need to figure out whether JobExecutor is an ABC or not.

> You are forcing JobExecutor to not being an ABC if you define a public constructor for it.

Where is that coming from?  Is that a Java thing?  In C++, an abstract base class can very well have a public ctor.  A C++ class is an ABC if it has at least one pure virtual function.  In Python...  well, anything goes.  How would an ABC for a `File` object work in Java - would all specializers which inherit/implement the ABC define their individual ctors which may differ (`ImageFile(image=...)`, `AudioFile(file=...)`)?  Either way, not buying that argument.

>>    Yeah, adaptors can inherit from a base class.
>
>Not "can", but "must" in strictly typed languages (notwithstanding contrived voodoo meant to bypass a mechanism for which a perfectly reasonable solution is already provided).

I can of course implement adaptors with a non-abstract base class, or no base class at all.  Even in strongly typed language.  But again, why are we discussing implementation details on the adaptor level, which is not part of the public API?  

> The specification specifies an API, which consists of classes with methods and properties. The understanding is that implementing libraries will have those classes and those methods and those properties. The understanding IS NOT that implementing libraries will ONLY have those classes and methods and properties. Again, we are deciding what goes into the PUBLIC API, not what goes into the implementing classes.

Ack, we only talk about the public API.

> So even though JobExecutor, as an ABC, may have a constructor, this does not imply that it needs to be specified by JPSI.

I expect of an portable API that *all* public method are specified.  Otherwise portability goes out of the window in a heartbeat.  If one implementation implements `je.submit_special()` and another implements `je.submit_extra()`, then what's the purpose of the spec?  A user needs to change code when switching implementation, so the spec becomes kind of useless.  Or am I missing the point you are trying to make?


> To the issue of portability:
>
>   1. I think we still haven't decided the extend of source-level compatibility we want to have

See [here](https://github.com/ExaWorks/job-api-spec/issues/27#issuecomment-723664165): I expect that at least a user can switch between implementations by importing  or including/relinking a different implementation, w/o other code changes.  I personally don't care about binary compatibility, or compatibility on the adaptor layer (if one exists in the implementation).  That's the minimal portability definition which seems to make sense.  

The discussion in that link above seemed to be somewhat resolved (even though not explicitly stated).  Can you agree to that type of source code portability? 


>   2. we can either leave the mechanism for instantiation open to library implementers, or we can specify one

Uhm, I kind of argue for the second one - just in case that got lost :-)


>   3. there are a number of options, such as factories, what SAGA is doing, or manual construction.

The public SAGA API is (generally) not using factories, but manual construction.  I said a couple of times: since the API defined in this document is supposed to be language independent, I don't care about c'tor vs. factory vs. initializer - I care that I can initialize the executor with LRM endpoint information.

>    You picked one of these solutions in this PR, implicitly answering both (2) and (3).

Again, and really, we had this a couple of times: I don't care about the mechanics of initialization - I care about being able to initialize with some LRM endpoint information.  If the PR does not fit the Java centric spec language, feel free to change - I *really* don't care.  The mechanism is up for the language binding.

>  I think we need to first discuss and come to an agreement on (2) and (3), because what you seem to have picked does not match what I or others I've talked to believe is the most common solution in this scenario.

I think this is what we do.


>  It is also fine to not have complete source-level compatibility as far as I'm concerned.

Uhm... what compatibility is then left?

hategan:
> > Are you saying that `File("/etc/rc.local")` and `File("/bin/ls")` is somewhat equivalent to `JobExecutor("slurm://")` and `JobExecutor("torque://")`?
> 
> Yes, absolutely. 

I would urge you to stop and think about this a bit. There is no difference in the code implementing `File` between `File("/etc/rc.local")` and `File("/bin/ls")`, but there is different code between a SLURM executor and a Torque one. 

> 
> > If JobExecutor is an ABC, it makes no sense to have a public constructor for it since it can never be called directly by user code. So, in order to figure out whether a constructor for JobExecutor makes sense, we need to figure out whether JobExecutor is an ABC or not.
> 
> > You are forcing JobExecutor to not being an ABC if you define a public constructor for it.
> 
> Where is that coming from? Is that a Java thing? In C++, an abstract base class can very well have a public ctor.

If it has parameters, you have to override it in both Java and C++ to be able to call it while constructing an instance of a derived class. No?

```C++
class Base {
    public:
        Base(int a) {}
        virtual void x() = 0;
}

class Derived {
    public:
        void x() {}
}

main() {
    Base *b = new Derived(1); //compiler error
}
```

Am I missing something?
So you must explicitly add a ctor in Derived. However, it need not have the same signature as the Base ctor:

```C++
class Derived {
    public:
        Derived(int a, int b) : Base(a) {}
        void x() {}
}

main() {
    Base *b = new Derived(1, 2); //all good
}
```

So the signature of the ABC ctor is irrelevant from a public API perspective. Nothing in C++ forces you to have the same signature for one or any of the Derived ctors. Conversely, if something is relevant from a public API perspective, which is implied if it's in the API spec, it cannot be an ABC ctor. But it is a ctor. So then the only thing that remains is that the class can't be an ABC. No?

And yeah, you can't do it in Java either. That, alone, should be disqualifying if we're going for a spec that should be implementable in commonly used OO languages.

> A C++ class is an ABC if it has at least one pure virtual function. In Python... well, anything goes. How would an ABC for a `File` object work in Java - would all specializers which inherit/implement the ABC define their individual ctors which may differ (`ImageFile(image=...)`, `AudioFile(file=...)`)? Either way, not buying that argument.

Yes. You need to specify constructors. There is an exception for default constructors if you don't have other constructors. But, again, it's almost the same in C++ as far as I can tell.

> 
> > > Yeah, adaptors can inherit from a base class.
> > 
> > 
> > Not "can", but "must" in strictly typed languages (notwithstanding contrived voodoo meant to bypass a mechanism for which a perfectly reasonable solution is already provided).
> 
> I can of course implement adaptors with a non-abstract base class, or no base class at all. Even in strongly typed language.

And be able to call them concurrently from client code without re-compiling in a way nicer and simpler than ABC/inheritance?

> But again, why are we discussing implementation details on the adaptor level, which is not part of the public API?

This is not an implementation detail. We must be able to use adapters interchangeably from client code without re-compiling said client code.

> 
> > The specification specifies an API, which consists of classes with methods and properties. The understanding is that implementing libraries will have those classes and those methods and those properties. The understanding IS NOT that implementing libraries will ONLY have those classes and methods and properties. Again, we are deciding what goes into the PUBLIC API, not what goes into the implementing classes.
> 
> Ack, we only talk about the public API.
> 
> > So even though JobExecutor, as an ABC, may have a constructor, this does not imply that it needs to be specified by JPSI.
> 
> I expect of an portable API that _all_ public method are specified. Otherwise portability goes out of the window in a heartbeat. If one implementation implements `je.submit_special()` and another implements `je.submit_extra()`, then what's the purpose of the spec? A user needs to change code when switching implementation, so the spec becomes kind of useless. Or am I missing the point you are trying to make?

See above around the C++ discussion. Are we maybe conflating the c++ notion of "public" with "public" in the sense of callable by or relevant to the user such that it needs to be in the spec?

> 
> > To the issue of portability:
> > 
> > 1. I think we still haven't decided the extend of source-level compatibility we want to have
> 
> See [here](https://github.com/ExaWorks/job-api-spec/issues/27#issuecomment-723664165): I expect that at least a user can switch between implementations by importing or including/relinking a different implementation, w/o other code changes. I personally don't care about binary compatibility, or compatibility on the adaptor layer (if one exists in the implementation). That's the minimal portability definition which seems to make sense.
> 
> The discussion in that link above seemed to be somewhat resolved (even though not explicitly stated). Can you agree to that type of source code portability?

It will be hard. We'll have to agree on lots of details, and we're not easily agreeing on details. And if we do come up with a rigid spec, we have zero guarantees that implementations will actually follow it. But let's say we have to agree on it with respect to this particular topic (i.e., points (2) and (3) below).

> 
> > 1. we can either leave the mechanism for instantiation open to library implementers, or we can specify one
> 
> Uhm, I kind of argue for the second one - just in case that got lost :-)

I agree with that.

> 
> > 1. there are a number of options, such as factories, what SAGA is doing, or manual construction.
> 
> The public SAGA API is (generally) not using factories, but manual construction. I said a couple of times: since the API defined in this document is supposed to be language independent, I don't care about c'tor vs. factory vs. initializer - I care that I can initialize the executor with LRM endpoint information.

There are two separate problems, I think:
1. picking between, say, SLURM and Torque adapters
2. passing an endpoint (or other configuration) to adapters

Let's finish the first for now. The other is related to "configuration" and "stateful/stateless", which I mentioned earlier. But let's solve one thing at a time.

> 
> > You picked one of these solutions in this PR, implicitly answering both (2) and (3).
> 
> Again, and really, we had this a couple of times: I don't care about the mechanics of initialization - I care about being able to initialize with some LRM endpoint information. If the PR does not fit the Java centric spec language, feel free to change - I _really_ don't care. The mechanism is up for the language binding.

I apologize for reading too much into this. It was my impression that you really wanted a constructor for JobExecutor to be used in the way it is used currently by the SAGA job.Service.

> 
> > I think we need to first discuss and come to an agreement on (2) and (3), because what you seem to have picked does not match what I or others I've talked to believe is the most common solution in this scenario.
> 
> I think this is what we do.

Ok. Can we agree that JobExecutor is an ABC that adapters are derived from and that we need to find a way to instantiate adapters (without passing a configuration at first, but we'll return to that)?

> 
> > It is also fine to not have complete source-level compatibility as far as I'm concerned.
> 
> Uhm... what compatibility is then left?

Having to make a few cosmetic edits here and there is much better than having to re-design my whole program because, say, the API I'm switching to is synchronous and the old one was async.

Ideally, I'd like to be able to say that we came with a spec that somehow makes all implementations be switchable by changing some import statements. But you should note that, should we achieve that with the spec, you and I, writing two different implementations of the spec, would come up with two things that can be swapped by changing import statement only. Do you think that's realistic? Also, do you think it will really be that common for projects to switch implementations of the JPSI spec that this is worth optimizing?

andre-merzky:
> > > Are you saying that `File("/etc/rc.local")` and `File("/bin/ls")` is somewhat equivalent to `JobExecutor("slurm://")` and `JobExecutor("torque://")`?
> > 
> > Yes, absolutely.
> 
> I would urge you to stop and think about this a bit. There is no difference in the code implementing `File` between `File("/etc/rc.local")` and `File("/bin/ls")`, but there is different code between a SLURM executor and a Torque one.

I think you missed the next sentence: `Take an example: I run two flux instances in userspace, and the same application wants to submit tasks to instance one or instance two.`  This is the same code.


> > > You are forcing JobExecutor to not being an ABC if you define a public constructor for it.
> > 
> > Where is that coming from? Is that a Java thing? In C++, an abstract base class can very well have a public ctor.
> 
> If it has parameters, you have to override it in both Java and C++ to be able to call it while constructing an instance of a derived class. No?

Yes, but that has nothing to do with ABC, that's how C++ inheritance always works.  Please don't shift the goalpost: your point was that an ABC does not have a public c'tor.  And, a final time, I don't care about ABC or not, I care about an API I want to implement.

> So the signature of the ABC ctor is irrelevant from a public API perspective. 

Not if it is specified in the document.  

> Nothing in C++ forces you to have the same signature for one or any of the Derived ctors.

Sure.  But equally, nothing in C++ stops me from calling the `jpsi.Executor` `jspi.FooBar` - only the API spec stops me from doing so.  What is your point?

> > > > Yeah, adaptors can inherit from a base class.
> > > 
> > > Not "can", but "must" in strictly typed languages (notwithstanding contrived voodoo meant to bypass a mechanism for which a perfectly reasonable solution is already provided).
> > 
> > I can of course implement adaptors with a non-abstract base class, or no base class at all. Even in strongly typed language.
> 
> And be able to call them concurrently from client code without re-compiling in a way nicer and simpler than ABC/inheritance?

(a) having ABC does not necessarily remove the need for recompiling; (b) recompiling can be avoided w/o ABC.   And in languages like Python that who ABC discussion is mute anyway.


> > But again, why are we discussing implementation details on the adaptor level, which is not part of the public API?
> 
> This is not an implementation detail. We must be able to use adapters interchangeably from client code without re-compiling said client code.

We seem to have different ideas about adaptors: for me, adaptor code and API are *not* necessarily part of the public API, as I stated multiple times.  And since you quote the SAGA code now and then, you will have realized that this is not the case in SAGA either: the adaptor API (CPI) is *not* exposed to the end user, and also differs from the public API.  And, again, if an implementation is adaptor based or not is an implementation detail, why are we discussing this?  I can implement J/PSI w/o adaptors, so what is your point?


> > I expect of an portable API that _all_ public method are specified. Otherwise portability goes out of the window in a heartbeat. If one implementation implements `je.submit_special()` and another implements `je.submit_extra()`, then what's the purpose of the spec? A user needs to change code when switching implementation, so the spec becomes kind of useless. Or am I missing the point you are trying to make?
> 
> See above around the C++ discussion. Are we maybe conflating the c++ notion of "public" with "public" in the sense of callable by or relevant to the user such that it needs to be in the spec?

With public I just mean the part which appears in the API documentation, and which the application is supposed to use.
 
> > > 1. I think we still haven't decided the extend of source-level compatibility we want to have
> > 
> > See [here](https://github.com/ExaWorks/job-api-spec/issues/27#issuecomment-723664165): I expect that at least a user can switch between implementations by importing or including/relinking a different implementation, w/o other code changes. I personally don't care about binary compatibility, or compatibility on the adaptor layer (if one exists in the implementation). That's the minimal portability definition which seems to make sense.
> > The discussion in that link above seemed to be somewhat resolved (even though not explicitly stated). Can you agree to that type of source code portability?
> 
> It will be hard.

What do you mean: hard to agree, or hard to achieve?

> And if we do come up with a rigid spec, we have zero guarantees that implementations will actually follow it.

I am lost - what does that mean?  If someone intents to implement J/PSI, why would they not follow the spec?  Otherwise they would not implement J/PSI?   

 
> > > 1. there are a number of options, such as factories, what SAGA is doing, or manual construction.
> > 
> > The public SAGA API is (generally) not using factories, but manual construction. I said a couple of times: since the API defined in this document is supposed to be language independent, I don't care about c'tor vs. factory vs. initializer - I care that I can initialize the executor with LRM endpoint information.
> 
> There are two separate problems, I think:
> 
>     1. picking between, say, SLURM and Torque adapters
> 
>     2. passing an endpoint (or other configuration) to adapters
> 
> 
> Let's finish the first for now. The other is related to "configuration" and "stateful/stateless", which I mentioned earlier. But let's solve one thing at a time.

Why?  The second one is just as important to our use case.  And no, it's not related to configuration and stateful/stateless, or at least not any different from the first part.


> > > I think we need to first discuss and come to an agreement on (2) and (3), because what you seem to have picked does not match what I or others I've talked to believe is the most common solution in this scenario.
> > 
> > I think this is what we do.
> 
> Ok. Can we agree that JobExecutor is an ABC that adapters are derived from and that we need to find a way to instantiate adapters (without passing a configuration at first, but we'll return to that)?

No.  We define an API, not a language binding.  If an implementation uses inheritance or not is up to the implementation, so the notion of ABC is not useful IMHO.  If an implementation uses adaptors or not is up to the implementation, so that discussion is not useful either IMHO.    


> > > It is also fine to not have complete source-level compatibility as far as I'm concerned.
> > 
> > Uhm... what compatibility is then left?
> 
> Having to make a few cosmetic edits here and there is much better than having to re-design my whole program because, say, the API I'm switching to is synchronous and the old one was async.
> 
> Ideally, I'd like to be able to say that we came with a spec that somehow makes all implementations be switchable by changing some import statements. But you should note that, should we achieve that with the spec, you and I, writing two different implementations of the spec, would come up with two things that can be swapped by changing import statement only. Do you think that's realistic?

If I would not have that ability, then the spec is not useful to our use case.  The whole purpose of that API is, for us, to avoid code like
```
if use slurm,v2: ...
if use slurm,v3: ...
elif use pbs: ...
elif: use flux
...
```

I MUST be able to use the same code against different backends.

> Also, do you think it will really be that common for projects to switch implementations of the JPSI spec that this is worth optimizing?

That is the central use case for us!  It is the *only* use case for us.  I need to run my code *unchanged* on Frontera (slurm), Summit(LSF) and Flux (userspace).  Why would we define an abstracting API otherwise?  From our intro:

```
Traditionally, job management is implemented on supercomputers by Local Resource Managers
(LRMs), such as PBS/Torque, SLURM, etc. To a first approximation, a job management API is
 understood as an abstraction layer on top of various LRMs.
```

But really, Mihael, lets stop, this is not a useful discussion.  We are repeating the same arguments.  I say: I need to initialize the executor.  You say ABC's don't force initializers.  I say I need to talk to two Fluxes, and that needs initialization.  You say, slurm/PBS implementations have different codes.  I say that doesn't mean different API.  Your say that adaptors must use ABC.  I say I don't care about adaptors, but the API.  and  yada yada yada.

I think I was explicit, really:
  - I MUST be able to talk to two Flux instances to make this API in any way useful to RCT.
  - I MUST be able to switch Slurm/PBS without code changes to make this useful to RCT.
  - I MUST be able to talk to Slurm and Flux from within the same code base, in the same process, at the same time.
  - I SHOULD be able to implement this API w/o ABC to make it a useful spec for RCT.

hategan:
> > > > Are you saying that `File("/etc/rc.local")` and `File("/bin/ls")` is somewhat equivalent to `JobExecutor("slurm://")` and `JobExecutor("torque://")`?
> > > 
> > > 
> > > Yes, absolutely.
> > 
> > 
> > I would urge you to stop and think about this a bit. There is no difference in the code implementing `File` between `File("/etc/rc.local")` and `File("/bin/ls")`, but there is different code between a SLURM executor and a Torque one.
> 
> I think you missed the next sentence: `Take an example: I run two flux instances in userspace, and the same application wants to submit tasks to instance one or instance two.` This is the same code.

There's A (instantiating different adaptors) and B (instantiating the same adaptor with different parameters) and they are being conflated. You seem to be arguing that something that is natural for B should apply to A.

> 
> > > > You are forcing JobExecutor to not being an ABC if you define a public constructor for it.
> > > 
> > > 
> > > Where is that coming from? Is that a Java thing? In C++, an abstract base class can very well have a public ctor.
> > 
> > 
> > If it has parameters, you have to override it in both Java and C++ to be able to call it while constructing an instance of a derived class. No?
> 
> Yes, but that has nothing to do with ABC, that's how C++ inheritance always works. Please don't shift the goalpost: your point was that an ABC does not have a public c'tor. And, a final time, I don't care about ABC or not, I care about an API I want to implement.

When I specify a constructor in an ABC, there is no guarantee that user code, constructing a derived class, can call the constructor of the derived class with the same signature as the one for the ABC. By contrast, when I declare a public method for an ABC, it is guaranteed that user code can call that method on a derived class.

My point was that there is no point in specifying a constructor for an ABC. That's the one and only goalpost.

We can, of course, specify a constructor for derived classes. But, if we do that, we must state so. We cannot simply specify a constructor for what can very well be an ABC and pretend that it implies that derived classes can be instantiated using that constructor.

> 
> > So the signature of the ABC ctor is irrelevant from a public API perspective.
> 
> Not if it is specified in the document.

See above. ABC constructors do not constrain the way in which derived classes are constructed by user code.

> 
> > Nothing in C++ forces you to have the same signature for one or any of the Derived ctors.
> 
> Sure. But equally, nothing in C++ stops me from calling the `jpsi.Executor` `jspi.FooBar` - only the API spec stops me from doing so. What is your point?

Specifying a constructor for an ABC in the spec does not imply that derived classes have that constructor. If we want derived classes to have a certain constructor, we must explicitly state that, not just write a constructor for an ABC.

> 
> > > > > Yeah, adaptors can inherit from a base class.
> > > > 
> > > > 
> > > > Not "can", but "must" in strictly typed languages (notwithstanding contrived voodoo meant to bypass a mechanism for which a perfectly reasonable solution is already provided).
> > > 
> > > 
> > > I can of course implement adaptors with a non-abstract base class, or no base class at all. Even in strongly typed language.
> > 
> > 
> > And be able to call them concurrently from client code without re-compiling in a way nicer and simpler than ABC/inheritance?
> 
> (a) having ABC does not necessarily remove the need for recompiling;

Calling ABC methods from client code allows you to use arbitrary subclasses of the ABC without re-compiling.

Andre, what's going on? Why are we arguing about how OOP works?

> (b) recompiling can be avoided w/o ABC. And in languages like Python that who ABC discussion is mute anyway.

Sure it can be avoided. But there is a perfectly reasonable mechanism that was invented just for that purpose, and it's ABCs.

Unless the spec is only for Python, let's not use statements that only apply to Python.

> 
> > > But again, why are we discussing implementation details on the adaptor level, which is not part of the public API?
> > 
> > 
> > This is not an implementation detail. We must be able to use adapters interchangeably from client code without re-compiling said client code.
> 
> We seem to have different ideas about adaptors: for me, adaptor code and API are _not_ necessarily part of the public API, as I stated multiple times. And since you quote the SAGA code now and then, you will have realized that this is not the case in SAGA either: the adaptor API (CPI) is _not_ exposed to the end user, and also differs from the public API. And, again, if an implementation is adaptor based or not is an implementation detail, why are we discussing this? I can implement J/PSI w/o adaptors, so what is your point?

I don't think we decided that adapter code is NOT part of the public API. That's SAGA, not necessarily JPSI.

> 
> > > I expect of an portable API that _all_ public method are specified. Otherwise portability goes out of the window in a heartbeat. If one implementation implements `je.submit_special()` and another implements `je.submit_extra()`, then what's the purpose of the spec? A user needs to change code when switching implementation, so the spec becomes kind of useless. Or am I missing the point you are trying to make?
> > 
> > 
> > See above around the C++ discussion. Are we maybe conflating the c++ notion of "public" with "public" in the sense of callable by or relevant to the user such that it needs to be in the spec?
> 
> With public I just mean the part which appears in the API documentation, and which the application is supposed to use.

So do I. That's why I said specifying constructors for ABCs in the spec is useless.

> 
> > > > 1. I think we still haven't decided the extend of source-level compatibility we want to have
> > > 
> > > 
> > > See [here](https://github.com/ExaWorks/job-api-spec/issues/27#issuecomment-723664165): I expect that at least a user can switch between implementations by importing or including/relinking a different implementation, w/o other code changes. I personally don't care about binary compatibility, or compatibility on the adaptor layer (if one exists in the implementation). That's the minimal portability definition which seems to make sense.
> > > The discussion in that link above seemed to be somewhat resolved (even though not explicitly stated). Can you agree to that type of source code portability?
> > 
> > 
> > It will be hard.
> 
> What do you mean: hard to agree, or hard to achieve?

The more we have to specify, the more we have to agree on. So they are connected.

> 
> > And if we do come up with a rigid spec, we have zero guarantees that implementations will actually follow it.
> 
> I am lost - what does that mean? If someone intents to implement J/PSI, why would they not follow the spec? Otherwise they would not implement J/PSI?

I don't know. You tell me. Why do you have, for example, `Executor.register_callback()` instead of `Executor.set_status_callback()` in the implementation? Or why is there a constructor to Specification that allows you to build it from a dict? Or why is Specification not called JobSpecification?

Are these liberties that we should expect implementations to be able to take? Clearly one can't simply switch imports if so.

> 
> > > > 1. there are a number of options, such as factories, what SAGA is doing, or manual construction.
> > > 
> > > 
> > > The public SAGA API is (generally) not using factories, but manual construction. I said a couple of times: since the API defined in this document is supposed to be language independent, I don't care about c'tor vs. factory vs. initializer - I care that I can initialize the executor with LRM endpoint information.
> > 
> > 
> > There are two separate problems, I think:
> > ```
> > 1. picking between, say, SLURM and Torque adapters
> > 
> > 2. passing an endpoint (or other configuration) to adapters
> > ```
> > 
> > 
> > Let's finish the first for now. The other is related to "configuration" and "stateful/stateless", which I mentioned earlier. But let's solve one thing at a time.
> 
> Why?

We break down complex problems into simpler ones and tackle one at a time to make the process manageable instead of this.

> The second one is just as important to our use case. And no, it's not related to configuration and stateful/stateless, or at least not any different from the first part.

Agreed that it's important. It is, however, different. Different code vs. different parameters.

> 
> > > > I think we need to first discuss and come to an agreement on (2) and (3), because what you seem to have picked does not match what I or others I've talked to believe is the most common solution in this scenario.
> > > 
> > > 
> > > I think this is what we do.
> > 
> > 
> > Ok. Can we agree that JobExecutor is an ABC that adapters are derived from and that we need to find a way to instantiate adapters (without passing a configuration at first, but we'll return to that)?
> 
> No. We define an API, not a language binding. If an implementation uses inheritance or not is up to the implementation, so the notion of ABC is not useful IMHO. If an implementation uses adaptors or not is up to the implementation, so that discussion is not useful either IMHO.

Ok. So we are not mandating that JobExecutor is an ABC.

Can we agree that an implementation CAN choose to see JobExecutor as an ABC and have adapters be subclasses of JobExecutor?

> 
> > > > It is also fine to not have complete source-level compatibility as far as I'm concerned.
> > > 
> > > 
> > > Uhm... what compatibility is then left?
> > 
> > 
> > Having to make a few cosmetic edits here and there is much better than having to re-design my whole program because, say, the API I'm switching to is synchronous and the old one was async.
> > Ideally, I'd like to be able to say that we came with a spec that somehow makes all implementations be switchable by changing some import statements. But you should note that, should we achieve that with the spec, you and I, writing two different implementations of the spec, would come up with two things that can be swapped by changing import statement only. Do you think that's realistic?
> 
> If I would not have that ability, then the spec is not useful to our use case. The whole purpose of that API is, for us, to avoid code like
> 
> ```
> if use slurm,v2: ...
> if use slurm,v3: ...
> elif use pbs: ...
> elif: use flux
> ...
> ```
> 
> I MUST be able to use the same code against different backends.

Agreed. We have two things:
A. Switching JPSI implementations (aka. the entire library)
B. Switching adapters (aka. SLURM vs. PBS)

I think you interpreted this part of the conversation to be about (B) when I was referring to (A).

> 
> > Also, do you think it will really be that common for projects to switch implementations of the JPSI spec that this is worth optimizing?
> 
> That is the central use case for us! It is the _only_ use case for us. I need to run my code _unchanged_ on Frontera (slurm), Summit(LSF) and Flux (userspace). Why would we define an abstracting API otherwise? From our intro:

See above. This ("implementations of the [whole] JPSI spec") is about (A) not (B).

> 
> ```
> Traditionally, job management is implemented on supercomputers by Local Resource Managers
> (LRMs), such as PBS/Torque, SLURM, etc. To a first approximation, a job management API is
>  understood as an abstraction layer on top of various LRMs.
> ```
> 
> But really, Mihael, lets stop, this is not a useful discussion. We are repeating the same arguments. I say: I need to initialize the executor. You say ABC's don't force initializers. I say I need to talk to two Fluxes, and that needs initialization. You say, slurm/PBS implementations have different codes. I say that doesn't mean different API. Your say that adaptors must use ABC. I say I don't care about adaptors, but the API. and yada yada yada.

We agree on most things. I don't claim that adaptors must use ABCs, but it is one of the obvious and simplest way to solve the problem.

> 
> I think I was explicit, really:
> 
> * I MUST be able to talk to two Flux instances to make this API in any way useful to RCT.

Perfect. We can talk about initializing that thing that talks to Flux to communicate to it/them the correct instance.

> * I MUST be able to switch Slurm/PBS without code changes to make this useful to RCT.

Also perfect. A URL passed to a frontend JobExecutor is not the only way to do this and we should not mandate that.

> * I MUST be able to talk to Slurm and Flux from within the same code base, in the same process, at the same time.

Agreed. Inheritance is the mechanism invented to address just that problem and would be the most obvious way to structure the API, but I think you disagree with that, so, if we want consensus, we want to not mandate it one way or another.

> * I SHOULD be able to implement this API w/o ABC to make it a useful spec for RCT.

Excellent. I should also be able to implement this API by considering JobExecutor to be an ABC.

SteVwonder:
Thanks @andre-merzky and @hategan for the vigorous discussions and pushing this forward.  Sorry I haven't chimed in. I think this has discussion has branched off enough times that each reply has so many sub-replies as to make following all the threads very very difficult.  I highly recommend we open issues for each independent topic and continue the discussions there. The different threads (as I see them) are:
- One big document versus multiple documents
- How much compatibility to have across implementations (API, ABI, etc)
  - Should implementations deviate from or provide "plus ups" over the standard
- How to specify the standard in a more language-agnostic way (and should we have language-specific API documents
- If we do language-specific APIs, should we specify a class hierarchy, ABCs, constructors, etc
- Whether or not executors should be provided URLs in Layer 0 (the original topic of this PR, which should be continued in #23

Did I miss any topics (I'm sure I did)?  Do you disagree with any of these identified topics?  If not, I'll go ahead and start opening separate issues.

andre-merzky:
> Did I miss any topics (I'm sure I did)? Do you disagree with any of these identified topics? If not, I'll go ahead and start opening separate issues.

Thanks for helping us to disentangle this thread @SteVwonder !  That's a good list I think, lets dig through this one by one before we return here...

PR #28, opened 2020-11-08
Fix callback example

SteVwonder:
The first commit is just fixing a typo.

The second commit might be controversial depending on if it was intentional or not to make the job status callback setters for `Executor` and `Job` different.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -328,7 +328,7 @@ asynchronous implementation can run multiple jobs in a single thread:
 ```java
 void runJobs() {
     jobsLeft = alljobs.size();
-    executor.addJobStatusCallback(new JobStatusCallback() {
+    executor.setJobStatusCallback(new JobStatusCallback() {
         jobStatusChanged(Job job, JobStatus status) {
             if (status.isTerminal()) {
                 jobsLeft--;


specification.md
```diff
@@ -775,7 +775,7 @@ complete.
 
 <a name="job-setstatuscallback"></a>
 ```java
-void setStatusCallback(JobStatusCallback? cb)
+void setJobStatusCallback(JobStatusCallback? cb)
```
hategan, line 778: Job.setJobStatusCallback seems a bit redundant, but I agree that it should be consistent.

andre-merzky (approved):
LGTM, thanks

specification.md
SteVwonder: Ok.  Are you suggesting we change both to `setStatusCallback`, or are you OK with this as is?  If the latter, can you resolve this convo and merge?

hategan: I think we should keep the job one as `setStatusCallback`. I am unsure about the executor one, and I don't think uniformity is important here (despite what I said earlier). But, if you do want uniformity, I'd `setStatusCallback` in both places makes more sense.

SteVwonder: Ok.  I'll just drop that commit from this PR and we can deal with consistency later if we need to.

PR #35, opened 2020-11-14
wording change

danielskatz:
opening the wording slightly to allow the possibility that the management itself is not fully on the exascale machine (perhaps remote)

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -196,7 +196,7 @@ exec/popen.
 ## Introduction
 
 The purpose of this document is to provide an analysis of the design and
-implementation issues of a job management API suitable for managing jobs that are on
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
I changed this one a bit, but looks good to me otherwise - thanks!

danielskatz:
Your change lost part of my meaning - in your version, both the managing and the jobs is on the exascale system.  In my version, just the jobs were, and the managing didn't have to be.

andre-merzky:
Oh, apologies, I see that now!  I fixed this in the main branch, it contains now your original wording.
  
Thanks, Andre.

PR #40, opened 2020-11-15
moved sync vs. async in an appendix

hategan:


andre-merzky (approved):
I would suggest to move the appendixes into separate files maybe?  Otherwise that looks good to me.

hategan:
> I would suggest to move the appendixes into separate files maybe? Otherwise that looks good to me.

The linking wouldn't be portable between GH and GH Pages. In GH, the link would have to be to the .md file, whereas in Pages you would either use no extension or .html.

However, I think this problem can be addressed with a bit of JavaScript.

Anyway, you need to decide whether you want this to be done now or we can merge the PR and move this into an issue.

andre-merzky:
> Anyway, you need to decide whether you want this to be done now or we can merge the PR and move this into an issue.

It was a suggestion -- as that, happy to leave this to your judgement :-)

hategan:
> > Anyway, you need to decide whether you want this to be done now or we can merge the PR and move this into an issue.
> 
> It was a suggestion -- as that, happy to leave this to your judgement :-)

Well, you are the one who needs to close the conversation if you are satisfied with the state of things and, if there is nothing else in the review, merge the PR.

andre-merzky:
> Well, you are the one who needs to close the conversation if you are satisfied with the state of things and, if there is nothing else in the review, merge the PR.

fair...

PR #44, opened 2020-11-17
Added some of the assumptions/goals to the intro

hategan:


specification.md
```diff
@@ -204,11 +204,36 @@ implementation of a job management API is a job management library. A
...
+
+The main motivation behind the present job management API is the ubiquity with
+which projects meant to simplify the process of doing science on compute
+clusters are forced to implement their own solution. The reason is simple. If
+any tool needs to be portable across multiple clusters (and it rarely makes
+sense to not want such portability), it must access the underlying cluster LRM
+in an abstract way; that is, it must use a job management API. The only stable
+job management API currently available is
+[SAGA](http://radical-cybertools.github.io/) [so, wait a minute, how do we
+justify not pushing SAGA forward?].
```
andre-merzky, line 221: I think I will need to leave others to clarify this.  One argument I heard on the call (but not one I would make, just repeating) is that SAGA, or at least the name, comes with a certain pre-notation / ballast, and that would at least partially justify to distinguish an Exaworks job API from it.

BTW: DRMAA also is an existing, stable API -- just not a widely implemented one (anymore).   As such, we should also motivate why we do not implement DRMAA instead of inventing our own API.

specification.md
```diff
@@ -204,11 +204,36 @@ implementation of a job management API is a job management library. A
...
+describing characteristics of the target cluster/resource are beyond the scope
+of this API. This is motivated, in part, by the fact that such functionality
+would push the complexity of the API into unmanageable territory, while,
+simultaneously being better suited for separate components.
+
+We take inspiration from a number of projects, some defunct, with overlapping
+scope, such as [Globus
+GRAM](https://en.wikipedia.org/wiki/Grid_resource_allocation_manager),
+[SAGA](http://radical-cybertools.github.io/), and
+[DRMAA](http://www.drmaa.org/).
```
andre-merzky, line 236: I would suggest to add DRMAA, but also Flux here.  We could also add POSIX FWIW (qsub etc): even though POSIX does not define an LRM API, it does define those command line tools which are widely considered the standard interface in many LRM implementations.

andre-merzky (changes requested):
I left two comments which I think should be addressed.  Thanks!

Issue #27, opened 2020-11-08
API Compatibility Across Implementations?


andre-merzky:
My assumption is this:
  - the document defines, in a language independent way, a set of classes and methods (the 'API') which abstracts job specification,  submission and management towards an execution backend.
  - we expect different language bindings to emerge for that API
  - we expect different implementations to emerge for those language bindings
  - the document is informed by existing bindings and implementations, but is not normative toward the language bindings nor towards the implementations,

I am not sure if my assumptions are valid - please discuss.  Thank you.

hategan:
I agree. I might have made a point, however, on defining degree of "normativity" in a slightly different way: should we expect reasonable source-code compatibility between different implementations?

I would say yes: that a user should have to make minimal changes/refactoring if switching between implementations, but that we do not mandate that no changes whatsoever be necessary.

For example, I think if we specify properties for a class, they should be translated into getters and setters in Java, properties in Python, but not naked dictionaries in either. That is, I would not like a client to have to change `job.executable = '/bin/sleep'` into `job['executable'] = '/bin/sleep'` when switching implementations (even if `job.executable = '/bin/sleep'` is internally implemented as `self._dict['executable'] = value`. In Java, this makes even more sense because the setters/getters allow more fine-grained type checking to occur, so, while it doesn't matter much in python, there is a material difference in Java (and strongly typed languages).

That said, it's totally not the end of the world if somebody goes ahead and implements things with dictionaries. We will have no official certification mechanism, no means to enforce any of these rules, and, to a large extent, we shouldn't want to: somebody may actually make a better version of this API (although, perhaps, that won't be by implementing everything with dictionaries). The single biggest value of this API at Layer 0 would be to have re-usable implementations of programmatic interfaces to LRMs that have some minimal standards of decency.

andre-merzky:
> should we expect reasonable source-code compatibility between different implementations?

Well, that is why we have an API, right?  In Python I would assume this:

```py
import flux.sagav2 as saga
je = saga.JobExecutor()
job = saga.Job()
je.submit(job)
...

import slurm.sagav2 as saga
je = saga.JobExecutor()
job = saga.Job()
je.submit(job)
...

I would assume this to be valid.  If the API is well defined, we can switch implementations.  Same for C (include different headers, link different library - but the application code remains unchanged) or any other language.

[...]

PR #47, opened 2020-11-18
added some examples in an appendix

hategan:
There is a TODO with ideas for examples, but I'm not quite able to address all of them directly. So I'll set this PR as a draft to discuss those issues. Specifically:

- [ ] do the same but use the ListException from the submit call to do
  the same

"the same" being ensuring that concurrent jobs are throttled to some maximum. I am assuming that whoever added this intended to track some "too many jobs in the queue" error from the LRM and act according to that. If so, it would be difficult to do this because we do not define a clear mechanism that could allow a clear way for a user to distinguish such an error from other errors. To state that differently, it's not impossible, but it isn't specified. One would have to rely on how the implementation happens to report errors of such type.

- [ ] Submit a malformed or unsatisfiable job, then check for the error
  and print it out

Does anybody remember what the intent was here?

- [ ] Construct a job that uses all the various "knobs" of the resource

I think this is rather generic. Can we be more specific?

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -60,6 +60,10 @@
 			- [Threaded Submission](#threaded-submission)
 			- [Asynchronous Networking](#asynchronous-networking)
 			- [Connection Multiplexing](#connection-multiplexing)
+		- [Appendix C - examples](#appendix-c-examples)
+			- [Submit and wait for N jobs](#submit-and-wait-for-n-jobs)
+			- [Run N jobs while throttling to M concurrent jobs](#run-n-jobs-while-throttling-to-m-concurrent-jobs)
+			- [Have N jobs compete in the queue and keep only the winner](#have-n-jobs-compete-in-the-queue-and-keep-only-the-winner)
 
 <!-- /TOC -->
 
@@ -106,6 +110,28 @@ string depending on programming  language
     - [x] Make async vs sync much shorter with a pointer to the full text in the
     appendix
 
+- [ ] add examples of how one would use this API (and please, if you have
+    any "how do you do x?", please add here)
+
+    - [x] Submit X jobs and wait for them to complete in order of submission
+    with the waitFor function
+
+    - [x] Submit X jobs but ensure only Y job are queued/running at a time
+    (rolling window implemented with the jobStatusCallback)
+
+    - [ ] do the same but use the ListException from the submit call to do
+    the same
+
+    - [x] Submit a job, wait for it to be queued, then cancel it, then wait
+    for it to complete/be killed.
+
+    - [ ] Submit a malformed or unsatisfiable job, then check for the error
+    and print it out
+
+    - [ ] Construct a job that uses all the various "knobs" of the resource
+    and job specifications (with some verbose comments thrown in)
+
+
 - [ ] Consider adding further exceptions to submit() in order to
 distinguish between EAGAIN types of errors and others.
 
@@ -135,27 +161,6 @@ with (they never quite get the launching part right).
     launchers
 
 - [ ] add a section giving an overview of the API components
-
-- [ ] add examples of how one would use this API (and please, if you have
-any "how do you do x?", please add here)
-
-    - Submit X jobs and wait for them to complete in order of submission
-    with the waitFor function
-
-    - Submit X jobs but ensure only Y job are queued/running at a time
-    (rolling window implemented with the jobStatusCallback)
-
-    - do the same but use the ListException from the submit call to do
-    the same
-
-    - Submit a job, wait for it to be queued, then cancel it, then wait
-    for it to complete/be killed.
-
-    - Submit a malformed or unsatisfiable job, then check for the error
-    and print it out
-
-    - Construct a job that uses all the various "knobs" of the resource
-    and job specifications (with some verbose comments thrown in)
 
 - [ ] add some clarification about the correspondence between JobSpec and
 exec/popen.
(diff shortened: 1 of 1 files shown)


hategan:
I am going to remove the draft part from this PR. I think it should be merged and then we can talk about other examples in a separate issue.

hategan:
> * Submit a malformed or unsatisfiable job, then check for the error
>   and print it out
> 
> Does anybody remember what the intent was here?

Intent was to use the exception types to distinguish between re-triable or not and print message to user if the latter.

hategan:
> * [ ]  Construct a job that uses all the various "knobs" of the resource
>   and job specifications (with some verbose comments thrown in)
> 
> I think this is rather generic. Can we be more specific?

Please add exotic use cases to this issue and will add examples as needed.

specification.md
```diff
@@ -1574,3 +1579,119 @@ to send over the connection. When using a single connection to transmit
...
+    spec.executable = '/bin/sleep'
+    spec.arguments = ['10']
+    job.specification = spec
+    return job
+
+jobs = []
+for i in range(N):
+    job = make_job()
+    jobs.append(job)
+    jex.submit(job)
```
andre-merzky, line 1613: Shouldn't we use the bulk submit?
```py
jobs = [make_job() for _ in range(N)]
jex.submit(jobs)
```

specification.md
```diff
@@ -1574,3 +1579,119 @@ to send over the connection. When using a single connection to transmit
...
+This Appendix contains some examples of how the API can be used. Unlike the
+specification language, the examples are in a hypothetical Python binding, which
+is expected to be a relatively frequently used binding.
+
+#### Submit and wait for N jobs
+
+This example shows how to submit `N` jobs and synchronously wait for them to
+complete.
+
+```python
```
andre-merzky, line 1596: Would you mind if I add a short note that the Python code used in examples is not normative and that Python bindings may deviate? :-)

specification.md
```diff
@@ -1574,3 +1579,119 @@ to send over the connection. When using a single connection to transmit
...
+import psij
+
+class ThrottledSubmitter:
+    def __init__(self):
+        self.jex = psij.JobExecutorFactory.get_instance('torque', '>= 0.2')
+        # keep track of completed jobs so that we can submit the rest
+        self.jex.set_status_callback(self.callback)
+        self.crt = 0
```
andre-merzky, line 1635: What does `crt` mean mean / refer to?

specification.md
```diff
@@ -1574,3 +1579,119 @@ to send over the connection. When using a single connection to transmit
...
+    job.specification = spec
+    return job
+
+jobs = []
+for i in range(N):
+    job = make_job()
+    jobs.append(job)
+    jex.submit(job)
+
+for i in range(N):
```
andre-merzky, line 1615: Hmm, I really think that we should make the full API bulk-compatible:
```py
jobs = [make_job() for _ in range(N)
jex.submit(jobs)
jex.wait(jobs)
```

andre-merzky (commented):
Thanks!

specification.md
hategan: See below

specification.md
hategan: I don't see where the degree of freedom is. The API should be constraining enough such that such notes should not be necessary. I mean I'm worried about sending the wrong message here, which is that the spec doesn't really matter: implementations should briefly read it and do whatever they want anyway.

specification.md
hategan: I'll rename.

specification.md
hategan: Perhaps I should more clearly state what I think about the bulk API: nothing that we discussed so far suggests a reason to have it except that it may make an efficient implementation easier to implement, while making the interface more complex. Everything else points to not having a bulk API. I do not think having a bulk API is, therefore, justified.

specification.md
andre-merzky: Well, we wanted to eventually have a separate binding specification (see #27), or at least that is how I remember our last discussion on this topic went?  Well, what I am really getting at is this: I would do things different in Python, but I don't want to muddy the waters on this PR by discussing how the Python code should look like, when the point is to add examples which demonstrate what functionality / semantics the API tries to cover.

specification.md
andre-merzky: Thanks for the clarity of the statement, that helps :-D  Let me follow suit:  I think bulks are essential on the API level.  Apart from that, they are also easy to use and make code simpler.

Seems we have some work to do :-P

Why do you think they make the interface more complex?  I am puzzled by that statement:
```py
jobs = [make_job() for _ in range(N)]
for job in jobs:
    jex.submit(job)

for job in jobs:
    job.wait()
```
vs
```py
jobs = [make_job() for _ in range(N)]
jex.submit(jobs)
jex.wait(jobs)
```

The loops are obvious, but the first also lacks symmetry (submit is on `jex`, wait is on `job`).  I mean, both are kinda simple of course - but I don't see how bulks add complexity.

Also: the point of bulks is not to make the implementation simpler (it does not), but to make an efficient implementation possible.

hategan: > Thanks for the clarity of the statement, that helps :-D Let me follow suit: I think bulks are essential on the API level. Apart from that, they are also easy to use and make code simpler.
> 
> Seems we have some work to do :-P

I say let's settle this once and for all... outside... in Issue #55 :)

> 
> Why do you think they make the interface more complex? 

Because every time the user needs to submit jobs, they need to decide whether to use the bulk or the non-bulk call. If the reason for having a bulk call is that implementations will possibly have better performance if the bulks are used, users will strive to use the bulk call even if they need to submit one job at a time. As I have stated before, we are bleeding implementation details into the API, and we are increasing the cognitive load on the users, but we do so without providing enough information so that the users can make an informed decision. Is it worth it to a user to queue single jobs only to use the bulk call? That can only be answered if the implementation is clear on how efficient the bulk call is vs. the non-bulk call. And it cannot really be answered in general, for all implementations. So, ultimately, our user will probably decide to err on the side of caution and always call the bulk API, implementing a queue if necessary to bulk up single jobs. But this implies that we are failing at design. If we know that this is very likely to happen, we should offset the effort from the user and do this bulking ourselves, so that the bulking code only gets written once. I mean, this is fundamentally what we are doing in this project: writing this API and adapters to that users don't have to implement their own every time they have to submit jobs in a portable way.

I'll leave bulk exception handling out, but that's an additional complication. It is pretty clear in the current API. The existence of the FaultDetail class is solely due to the bulk exceptions.

>I am puzzled by that statement:
> 
> ```python
> jobs = [make_job() for _ in range(N)]
> for job in jobs:
>     jex.submit(job)
> 
> for job in jobs:
>     job.wait()
> ```
> 
> vs
> 
> ```python
> jobs = [make_job() for _ in range(N)]
> jex.submit(jobs)
> jex.wait(jobs)
> ```
> 
> The loops are obvious, but the first also lacks symmetry (submit is on `jex`, wait is on `job`). I mean, both are kinda simple of course - but I don't see how bulks add complexity.

I would like to point out that one can have this argument for almost everything: add bulk versions so that users don't have to write for loops.

> 
> Also: the point of bulks is not to make the implementation simpler (it does not), but to make an efficient implementation possible.

I think the bulk submission appendix goes over this. One can queue jobs and periodically drain the queue into an internal bulk submit without exposing that to the user. But there are other ways to do things that can lead to even better performance. An efficient implementation is possible with or without the public bulk methods. 

The way bulks can make an implementation easier is by avoiding the aforementioned queue. But it's doing so at the expense of the user who might now have to do write that code themselves.

specification.md
hategan: I think my initial reply wasn't quite useful. Perhaps it would be more useful to see what you had in mind, and maybe just use that style for the examples.

specification.md
```diff
@@ -1574,3 +1579,119 @@ to send over the connection. When using a single connection to transmit
...
 Additionally, it becomes feasible to tune buffer sizes in order to
 optimize the throughput of connections to particular services.
+
+
+
+### Appendix C - examples
+
+This Appendix contains some examples of how the API can be used. Unlike the
+specification language, the examples are in a hypothetical Python binding, which
+is expected to be a relatively frequently used binding.
```
andre-merzky, line 1589: Thanks for adding that.

specification.md
andre-merzky: Resolving since we moved discussion to #55

andre-merzky (approved):
We moved the bulk discussion out, so this is ok to merge now.

specification.md
hategan: I don't think I did. That is, I think it might have been there from the start.

But I still think it's useful to see what you had in mind w.r.t. how an implementation of the spec might look like. Maybe we should open a new issue for that. Might help tell us if we're over or under specifying.

andre-merzky: Then I missed it on the first read :-)  

> But I still think it's useful to see what you had in mind w.r.t. how an implementation of the spec might look like.

Have a look!  https://github.com/ExaWorks/prototype-saga/tree/feature/radical_prototype/radical

PR #49, opened 2020-11-19
Removed timeline from STATUS section, since we now keep such records …

hategan:
…in Git, and replaced the STATUS section title with a simple description of where things are (currently "early draft") so that it's clear what a reader can expect.


andre-merzky (approved):
makes sense - thanks.

PR #63, opened 2020-12-11
Removed TODO items that are done. There is no point in keeping them a…

hategan:
…round except in the git history.


andre-merzky (approved):
Makes sense.

Issue #55, opened 2020-11-21
To bulk or not to bulk


hategan:
This issue is here so that we can decide whether to have bulk operations in the API (or not) and, possibly, the extent to which we should have them (i.e., what calls are to also have a bulk version).

There is a long history in the project of scattered discussions on this topic. This issue is meant to gather them in one place.

The starting point was the `submit(Job)` method. It was argued that a `submit(Job[])` bulk version could be useful. The [Bulk Submission Appendix](http://exaworks.org/job-api-spec/specification#appendix-c-bulk-submission) also goes over the details and spells out why bulk calls are unnecessary.

A related discussion is now happening in PR #47. We should move that discussion here.

andre-merzky:
Moving parts of the discussion:



>>    Why do you think they make the interface more complex?

> Because every time the user needs to submit jobs, they need to decide whether to use the bulk or the non-bulk call. If the reason for having a bulk call is that implementations will possibly have better performance if the bulks are used, users will strive to use the bulk call even if they need to submit one job at a time. As I have stated before, we are bleeding implementation details into the API, and we are increasing the cognitive load on the users, but we do so without providing enough information so that the users can make an informed decision. Is it worth it to a user to queue single jobs only to use the bulk call? That can only be answered if the implementation is clear on how efficient the bulk call is vs. the non-bulk call. And it cannot really be answered in general, for all implementations. So, ultimately, our user will probably decide to err on the side of caution and always call the bulk API, implementing a queue if necessary to bulk up single jobs. But this implies that we are failing at design. If we know that this is very likely to happen, we should offset the effort from the user and do this bulking ourselves, so that the bulking code only gets written once. I mean, this is fundamentally what we are doing in this project: writing stuff so that users don't have to each write their own version of said stuff.

> I'll leave bulk exception handling out, but that's an additional complication. It is pretty clear in the current API. The existence of the FaultDetail class is solely due to the bulk exceptions.

>>    I am puzzled by that statement:

```py
    jobs = [make_job() for _ in range(N)]
    for job in jobs:
        jex.submit(job)

    for job in jobs:
        job.wait()
```
>>    vs
```py
    jobs = [make_job() for _ in range(N)]
    jex.submit(jobs)
    jex.wait(jobs)
```
>>    The loops are obvious, but the first also lacks symmetry (submit is on jex, wait is on job). I mean, both are kinda simple of course - but I don't see how bulks add complexity.
>
[...]

PR #66, opened 2020-12-12
Removed the bulk submit method. This is related to issue #55.

hategan:
Additionally:
- removed todo item with example that would have used the bulk submit() call
- fixed a minor formatting/wrapping issue (@@ 156)
- simplified API by merging FaultDetail into SubmitException


andre-merzky (approved):
LGTM, thanks

Issue #31, opened 2020-11-12
Wish list from Balsam


hategan:
- dynamic attrs
    - on Theta need to pass specific parameter to get SSD storage - we need to investigate the details
    - flags for singularity

- pre-jobs for setting things up vs. wrapper scripts
    - wrapper scripts breaking abstraction
    - but perhaps so does populating pre-job info
    - do we abstract beyond job submission (e.g., install location of various tools, etc.)

kylechard:
List provided by @masalim2 (thanks Misha).

andre-merzky:
> pre-jobs for setting things up vs. wrapper scripts

@SteVwonder , @hategan : time ran out on yesterday's call to discuss this, but I want to voice an opinion here, or an perspective if you so will:  we have been using various variations and combinations of environment setup, and it might be useful to consider the lessons learnt.  

Our user like `pre_exec` and `post_exec`, as it provides tight control over the job environment.  Typical `pre_exec` are combined of commands like 

1. `module load foo`
3. `export LD_LIBRARY_PATH=$FOO_DIR/lib/special`
2. `. /some/ve/bin/activate`
4. `cp -r /shared_fs/large/data/ /local_fs/ssd/my_cache/`
5. `export WORK_ITEM=/local_fs/ssd/my_cache/$SLURM_STEP_ID.dat`

 `pre_exec` is easily misunderstood and abused.  Just for fun: before reading further, try to find the problems with the above `pre_exec` :-)

Problems:

 * [1] and [3] are *really* heave on the shared FS ([3] in particular), and only should be run once per job.  
 * [2] depends on [1] (uses `$FOO_DIR`).  
 * [4] needs to run once per node, as it populates node local cache - unless care is taken.  It is expensive if all ranks (even the ones on the same node) run [4]
 * [5] needs to run once per rank, as it depends on the rank
 * [5] also depends on the launcher: `$SLURM_STEP_ID` is srun specific, this would break if we launch with Flux etc.

Getting this right is complicated, getting the user educated is hard (but possible).  

At the moment, we are en route to separate `pre_exec` into two parts: generic environment setup (once per job), and rank specific initialization.  We cover node local caching via our staging directives, and we provide well defined `RP_*` env variables which are independent from the launcher (like, `RP_TASK_ID` etc.).  So these problems *can* be resolved, but it needs some care.  A very large percentage of user tickets and problems are related to `pre_exec`.

[...]

PR #67, opened 2020-12-12
Added back custom attributes to JobAttributes. This partially address…

hategan:
…es #31.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -88,14 +88,6 @@
     - [ ] Construct a job that uses all the various "knobs" of the resource
     and job specifications (with some verbose comments thrown in)
 
-<<<<<<< HEAD
-=======
-- [x] add ability to have custom attributes which could be passed to the
-underlying LRM (aka. dynamic attributes).
-
-- [ ] "canceled" or "cancelled"?
->>>>>>> 4b91b35... Added custom attributes to JobAttributes
-
 - [ ] Consider adding further exceptions to submit() in order to
 distinguish between EAGAIN types of errors and others.
 
--- a/specification.md
+++ b/specification.md
@@ -88,6 +88,13 @@
     - [ ] Construct a job that uses all the various "knobs" of the resource
     and job specifications (with some verbose comments thrown in)
 
+<<<<<<< HEAD
+=======
+- [x] add ability to have custom attributes which could be passed to the
+underlying LRM (aka. dynamic attributes).
+
+- [ ] "canceled" or "cancelled"?
+>>>>>>> 4b91b35... Added custom attributes to JobAttributes
 
 - [ ] Consider adding further exceptions to submit() in order to
 distinguish between EAGAIN types of errors and others.
@@ -95,9 +102,6 @@ distinguish between EAGAIN types of errors and others.
 - [ ] think more about env var expansion in arguments and other places.
 The important issue is how much of a burden this is on implementations if
 we mandate it.
-
-- [ ] add ability to have custom attributes which could be passed to the
-underlying LRM (aka. dynamic attributes).
 
 - [ ] we need to go through the resource spec; many common things
 supported by other JM APIs are not supported by Flux Jobspec V1, such as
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -88,6 +88,13 @@
     - [ ] Construct a job that uses all the various "knobs" of the resource
     and job specifications (with some verbose comments thrown in)
 
+<<<<<<< HEAD
```
andre-merzky, line 91: This looks conflicted...

specification.md
```diff
@@ -1063,6 +1067,19 @@ resources reserved through the advanced reservation represented by this
...
+Object? getCustomAttribute(String name);
+```
+
+Allows setting/querying of custom attributes. Implementations should use a
+dictionary to implement the custom attributes. Additionally, implementations are
+encouraged to make sensible decisions on whether to store some or all of the
+fixed attributes in the same dictionary or not. It is, therefore, entirely
+possible for `getCustomAttribute("duration")` to return a value passed earlier
+to `setDuration()`, although the specific custom attribute name need not be
+`"duration"`.
```
andre-merzky, line 1083: I would suggest not to speciofy the data storage type (dictionary) - the API allows any internal data storage, and I don't see a benefit in limiting that freedom.

As for the overlap with defined attributes, I see your point, but wonder if it wouldn't be easier for the user if no such overlap exists?  Because it implies also the inverted case: that the setting of a special attribute changes one of the other specified attribute's values (such as duration).  That sounds like surprising behaviour, doesn't it?  I also would worry to further impact portability, as it basically means that applications could code directly against special attributes w/o caring about the standardized ones - then why do we specify `duration` in the first place?

andre-merzky (commented):
Request for clarification - thanks!

specification.md
hategan: Not sure what happened there. Should be fixed now.

specification.md
hategan: My worry is that, without specifying that, the whole thing could be interpreted to require storage of the two in different places, which may affect the freedom of implementations to make certain optimizations. Basically, this is saying that we don't mandate it one way or the other and implementations are free to pick whichever solution they want.

But I see your point when it comes to the "custom" label on some of the attributes. That word isn't strictly necessary in the API and setAttribute(name, value)/getAttribute(name) are perfectly fine with me. Do you think removing "custom" would be a better choice?

andre-merzky: Is `Implementations are encouraged to make sensible decisions...` a useful statement?  I mean, we can make that statement on every single line of  code ;-)

As for the labels, I would suggest the following:
```
An implementation is not allowed to support custom attributes which are semantically equivalent to the attributes statically defined above (`Duration`, `QueueName`, `ProjectName`, `ReservationID`).  This removes potential confusion, and forces use of the proper API attributes.

hategan: > This removes potential confusion, and forces use of the proper API attributes.

Except that isn't what I want. I want the implementations to have the flexibility of doing that, too. For example, I want it to be possible to implement ALL attributes as a single dictionary and, for languages that support it, also have type-safe methods for putting/getting specific known attributes.

andre-merzky: > Except that isn't what I want. I want the implementations to have the flexibility of doing that, too. For example, I want it to be possible to implement ALL attributes as a single dictionary and, for languages that support it, also have type-safe methods for putting/getting specific known attributes.

That's a language binding issue.  In Python for example, both defined and custom attributes can be exposed as class properties, either one can have type checks or not, and the constructor can still have a dictionary initializer.  For languages which don't have that, I am not sure that having different, potentially conflicting routes to set a well defined attribute would be good.

I won't block the PR about that issue, I think you heard my point by now.

hategan: > I am not sure that having different, potentially conflicting routes to set a well defined attribute would be good.

It is somewhat common, in statically typed languages, to have properties with potentially unknown names be implemented using dictionaries while also providing typed accessor methods for the subset of properties that is known. I want to make it clear that the spec does not prevent such an implementation from happening.

I think I still want to consider renaming setCustomAttribute/getCustomAttribute to setAttribute/getAttribute. What do you think?

SteVwonder: > Because it implies also the inverted case: that the setting of a special attribute changes one of the other specified attribute's values (such as duration). That sounds like surprising behaviour, doesn't it? I also would worry to further impact portability, as it basically means that applications could code directly against special attributes w/o caring about the standardized ones - then why do we specify duration in the first place?

> For languages which don't have that, I am not sure that having different, potentially conflicting routes to set a well defined attribute would be good.

Just wanted to chime in to say that I agree.  I like the idea of the custom attributes but not as an alternate way to set/get values that we already have interfaces for.  Seems like this is asking for incompatibility across implementations and surprising behavior to users.

Issue #41, opened 2020-11-17
Accomodating short-lived clients


ejfried:
Some jobs remain in queue and/or run for days. It may not be possible (or allowed)  for a client program to stay up for the whole time, so there should be a mechanism to submit, query, and reap jobs from multiple distinct executions of a client program, spread over time and/or space.  I can think of two ways to do this, either:

1. The JobExecutor could have  query methods returning Job objects,  or
2. Job could have a constructor or a factory method to create live Job objects from the Job id.

andre-merzky:
I support this request, and would suggest to add:

```py
JobExecutor.list() -- returns list of job IDs known to the executor (qstat -u <user_id>
JobExecutor.get(uids) -- returns a list of job instances for the give job IDs
```

hategan:
I also am on board with the idea, but I would propose methods closer to the original comment:

`Job JobExecutor.attach(String nativeId)`. One would then query the status of the job using Job.getStatus().

I would also suggest making the `nativeId` a first class citizen, by adding `Job.getNativeId()`, which would be populated when the job makes it to a LRM queue.

Depending on what we ultimately decide about bulk calls (i.e., push this into the implementation and leave the API alone or not) we may also do `Job[] JobExecutor.attach(String[] nativeIds)` in order to avoid invoking `qstat` for each job being attached. We may have to augment the state model with an ATTACHED state, so that proper notification when an async attach operation completes is provided to the client.

andre-merzky:
> We may have to augment the state model with an ATTACHED state, so that proper notification when an async attach operation completes is provided to the client.

Oh, I don't think that would work: how would a job which is active differ when it is submitted and changes to `ACTIVE` and when I give a job ID and find it active?  It's still an active job.  Moreover, two clients looking at the same job would find it in different states (like, the original submitting application has `ACTIVE`, and a portal watching progress sees `ATTACHED`).

I think a state name should not reflect state history, in the sense it does not matter how the job instance got into that state, it matters that the job *is* in that state.

andre-merzky:
naming: `attach` that works well for me.  I am also on board to have an explicit query for the native ID.

> in order to avoid invoking qstat for each job being attached

`qstat` output can be cached, it will not be required to invoke it for every new job.  Having said that, I am all for bulk-enabling the full interface (where sensible) instead of only parts of it.

hategan:
[...]

PR #68, opened 2020-12-15
adress issue #69 (`list()` and `attach()`)

### LINKED ISSUE #69: Add a `executor.list()` method on the executor
This is related to PR #68 

The API includes an `executor.attach()` method which can be used to re-attach to a running (or at least known) job instance.  However, the API misses means to list known job instances.

Add an `executor.list()` method to return the IDs of known job instances which the API can attach to.

**Describe alternatives you've considered**
IDs can be retrieved out of band, which seems an impractical solution to me.

**Additional context**
Independent from the `list()` proposal above:
The API handles two types of job IDs: the native job ID as assigned by the underlying LRM system, and the native job ID as assigned by the J/PSI implementation.  Those IDs may or may not be the same.  `list()` should return the native IDs of the job.  Further, the document should clarify the scope of J/PSI IDs.  For example, if one attaches two job instances to the same native job ID, will that result in the same J/PSI ID?  If not, what is the exact purpose of the J/PSI ID?  If yes, can the native ID and the J/PSI ID be used interchangeably?

andre-merzky:
This addresses the proposal from #41:

  -  job.getNativeId()
  -  jobExecutor.attach(nativeID) -> job
  -  jobExecutor.list() -> list of nativeIDs

While writing the text, I wondered if we should also allow to attach to J/PSI IDs.  I mean, it would be kind of strange that we can attach to IDs created by the external LRM, but not to IDs created by our very own implementation?  That has some implications though, like that the implementation needs to separate the ID namespace from the LRM's native ID namespace - but that is not much of a problem.

Closes #69


hategan:
I think list() is not needed to address short-lived clients, so it is my suggestion that we keep/discuss list() separately.

specification.md
```diff
@@ -421,6 +421,35 @@ an individual job status callback or not. To remove the callback, set it
...
+Job attach(String jobID)
+```
+
+Return a job which is attached to the backend job identified by the specified
+ID.  The returned job instance MAY have an `UNKNOWN` state until the
+implementation asynchronously obtained information about the actual job state.
+The method MUST raise an `InvalidJobException` if the job ID cannot be used to
+uniquely identify a job.  The `jobID` can either refer to the ID returned by
+`Job.getID()`, which represents the ID used by the J/PSI implementation, or to
+the native job ID used by the LRM backend.
```
SteVwonder, line 451: > The `jobID` can either refer to the ID returned by `Job.getID()`, which represents the ID used by the J/PSI implementation

Is J/PSI assumed to be  stateless (that was an implicit assumption I just realized I was making)?  If so, how do you envision this functionality could be implemented?  If the J/PSI client stores some J/PSI ids to a file, restarts, and then tries to re-attach to the J/PSI ids in that file, how is the J/PSI library/runtime supposed to connect the old J/PSI ids to the underlying LRM's IDs?  Hopefully that question makes sense.

hategan: Right. We should perhaps also think forward on this particular issue, such as when talking to a remote service, in which case we need to also store and endpoint together with the LRM ID.

andre-merzky: Throwing this in here as reference points (not advocating one or the other):
  * DRMAA uses the underlying LRM job ID.  A job is created by the equivalent of `executor.submit()` and only returned once it has a job ID
  * SAGA uses the form `[LRM_ID]-[JOB_ID]`, where `LRM_ID` is the url to contact the LRM, and `JOB_ID` is the native job ID in that LRM instance.  SAGA jobs can have an undefined ID until the LRM assigns one to the job,
  
I am a bit worried that having two different job IDs may be confusing.  

Stateless: I did not make that assumption, but it is tempting.  But indeed, it makes consistent mapping to native IDs a bit difficult.  

> ... remote service, in which case we need to also store and endpoint together with the LRM ID.

Our `attach` method is on the Executor, and to get the executor, you would need to know the underlying LRM ID, so the native job IDs would be presumably unique in the scope of that LRM.  So yes, one needs both pieces of information, but it does not necessarily be part of the job ID.

andre-merzky:
I just realized that this PR misses to repsond on a comment from @hategan :

> returning a job is probably a bad idea if you want to add a callback. It would lead to possible race conditions. So maybe `JobExecutor.attach(Job job, String nativeId)` would be a better choice, such that client code can add the callback before invoking attach().

We did not come to a conclusion in the ticket.

specification.md
SteVwonder: > I am a bit worried that having two different job IDs may be confusing.

I share that worry.

> SAGA uses the form [LRM_ID]-[JOB_ID], where LRM_ID is the url to contact the LRM, and JOB_ID is the native job ID in that LRM instance. SAGA jobs can have an undefined ID until the LRM assigns one to the job,

That makes sense.  I would be fine with requiring as an argument to `attach` either `[LRM_ID]` or `[LRM_ID]-[JOB_ID]`.   I have reservations about adding to the standard the ability to just take in a `JOB_ID` without some other stated mechanism in the standard for recovering the underlying `LRM_ID` from a `JOB_ID` after a restart of the J/PSI client.  There are a bunch of ways we could do that, and I think we should open an issue to discuss that further.

SteVwonder:
It seems we have a few areas that require some further discussion, but we also might have areas of consensus.  In the name of getting things we agree on merged sooner rather than later, what do you all think of this:

- Open a separate issue for `list()`
- Open a separate issue for mapping `JOB_ID` -> `LRM_ID` after a restart
- Take out those two pieces (`list()` and `attach(job_id)`) from this PR and merge what we have consensus on: `getNativeID()` and `attach(native_id)`

@hategan : do you have any feedback or reservations related to the items in the last bullet (`getNativeID()` and `attach(native_id)`)?
@andre-merzky : would you be ok with getting the items in bullet 3 merged now and discussing 1 & 2 in other threads?

SteVwonder:
> We did not come to a conclusion in the ticket.

Ooops.  Sorry. I totally missed that. So I guess we need to iterate on the `attach` design a bit more then anyways.

SteVwonder:
> returning a job is probably a bad idea if you want to add a callback. It would lead to possible race conditions. So maybe JobExecutor.attach(Job job, String nativeId) would be a better choice, such that client code can add the callback before invoking attach().

That's a good point.  I'm guessing we would then need to expose a `Job` constructor that takes no arguments and builds an empty `Job` in the `NEW` state?  That's fine with me; just something to remember as we discuss constructors in #75 and elsewhere. It also opens the question as to what `getId` should respond when called on a "fresh" `Job`.  Will the `ID` returned change after a successful `attach`?

hategan:
> * Take out those two pieces (`list()` and `attach(job_id)`) from this PR and merge what we have consensus on: `getNativeID()` and `attach(native_id)`
> 
> @hategan : do you have any feedback or reservations related to the items in the last bullet (`getNativeID()` and `attach(native_id)`)?

I am on board with the general idea and I believe this has always been my preferred solution. I only object to mixing camel case and underscores :)

But I will add my two or so cents on the ID issue, some of which may have been made elsewhere:
- `job.id` is meant to be at least unique for a given client process
    - the basic idea being that it should be usable as a key in a hashtable of jobs, but not necessarily as a primary key in a db of all jobs run on a machine
    - we may want to discuss whether we want to make this globally unique (GUID/UUID), which would address the db pkey issue
- a native id may or may not be unique for a given client process
    - we can probably make it unique by combining it with LRM type and/or endpoint
    - speaking of endpoint, we should probably make native ids opaque (but serializable) so that they can support all necessary information to re-attach in both Layer 0 and 1.
- `job.id` is defined even before the job is submitted to a LRM; I believe it would problematic to not have an ID at all for a job until it gets queued, since, e.g., in a workflow system, there are many potential steps after the job is created and before the job is queued by the LRM.

specification.md
```diff
@@ -440,6 +440,35 @@ an individual job status callback or not. To remove the callback, set it
 to `null`.
 
 
+<a name="jobexecutor-list"></a>
+```java
+List<String> list()
+```
+
+Return a list of job IDs which are known to this executor instance.  The
```
SteVwonder, line 448:
```suggestion
Return a list of native job IDs which are known to this executor instance.  The
```

Do we want to call out explicitly that it is native IDs that are returned? Have we defined what a native ID is in the spec yet?

specification.md
```diff
@@ -440,6 +440,35 @@ an individual job status callback or not. To remove the callback, set it
...
+returned list MAY contain IDs of jobs which were not submitted via this
+instance, and MAY NOT contain IDs of jobs which have been submitted by this
```
SteVwonder, line 450:
```suggestion
instance, and MAY be missing IDs of jobs which have been submitted by this
```

The `MAY NOT` initially read to me like "the list IS NOT ALLOWED to contain IDs of jobs which have been submitted by this instance"

specification.md
```diff
@@ -440,6 +440,35 @@ an individual job status callback or not. To remove the callback, set it
...
+instance, but are finalized and purged already.  IDs for any job which has been
```
SteVwonder, line 451:
```suggestion
instance but are finalized and purged already.  IDs for any job which has been
```

SteVwonder (commented):
I think the `attach` stuff needs to be dropped from this PR since it was moved to #81.  Other than that, I think the `list` semantics here align pretty well with the latest discussion in #69 and on the phone.

andre-merzky:
The branch should now reflect the (many) discussions we had on the topic.  Since we agreed on including both `list()` and `attach()`, I recombined the two separate pull requests into this PR.

SteVwonder (approved):
LGTM! Thanks @andre-merzky!

specification.md
```diff
@@ -547,7 +595,20 @@ Returns this job's ID. The ID is assigned automatically by the
 implementation when the Job object is constructed. The ID is guaranteed
 to be unique on the client machine. The ID does not have to match the ID
 of the underlying LRM job, but is used to identify `Job` instances as
-seen by a client application.
+seen by a client application.  The ID can be used to later re-attach to the
+job with `Job j = executor.attach(native_job_id)`.
```
hategan, line 599: This is inconsistent with the signature above. I would suggest simply linking to the relevant method; it will also make it more robust in the future should the signature change again.

specification.md
```diff
@@ -547,7 +595,20 @@ Returns this job's ID. The ID is assigned automatically by the
...
+<a name="job-getnativeid"></a>
+```java
+String? getNativeId()
+```
+
+Returns this job's native ID as assigned by the underlying LRM.  The ID will
+only be available once the job has entered the `QUEUED` state - the returned
+value will be `null` otherwise.  The returned ID can be used to communicate
+with the LRM out-of-band, and also to later re-attach to the job with
+`Job j = executor.attach(native_job_id)`.
```
hategan, line 611: See above comment.

specification.md
```diff
@@ -454,6 +454,54 @@ to `null`.
...
+
+
+<a name="jobexecutor-attach"></a>
+```java
+Job attach(Job job, String nativeJobID)
+```
+
+This method will accept a job instance in `NEW` state and a native job ID.  The
+executor will attach the Job instance to the backend job identified by that
+native (backend) job ID.  The method will return immediately, the `JobExecutor`
```
hategan, line 492: That's a bit of a run-on. Perhaps "... will return immediately and the..."

specification.md
```diff
@@ -454,6 +454,54 @@ to `null`.
...
+will collect job state and meta data asynchronously.  A callback registered on
+the Job MUST NOT fire before the implementation completed the attachement -- at
+that point the job will have a valid `JobStatus`.  If the
+implementation is not able to attach the job (because it cannot verify
+the job ID, or the job information has been purged, etc), the job will
+be moved into `FAILED` state.
+
+The method will raise an [`InvalidJobException`](#invalidjobexception)
+if the passed job is not in `NEW` state.  Any job status, resource spec
+etc. which are set on the passed `Job` instance will be discarded.
```
hategan, line 502: Sorry for nitpicking, but should we mandate that this be so or warn that this could be so?

specification.md
```diff
@@ -454,6 +454,54 @@ to `null`.
...
+The returned IDs can be used to re-attach a `Job` instance to the backend job
+via the `executor.attach(job, nativeJobID)` call.  This implies that the call
+SHALL only return those IDs to which the callee can attach under currently used
+authorization.
+
+
+
+<a name="jobexecutor-attach"></a>
+```java
+Job attach(Job job, String nativeJobID)
```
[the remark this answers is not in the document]
andre-merzky: No, it does not - that's a leftover from a previous iteration, thanks for catching it.!

specification.md
andre-merzky: Thanks, fixed.

specification.md
andre-merzky: dito

specification.md
andre-merzky: Done

specification.md
andre-merzky: IMHO we should mandate, and I changed the text: MUST raise, MUST discard.  If we don't discard any conflicting attributes, we end up with an inconsistent set of information which may change over time, which is not good.  We could also mandate that the application did not set any attributes on the Job instance before calling - that would work for me also and might be cleaner?

andre-merzky: @hategan : ping

hategan: Sorry. Yeah, I see your point.

I'm considering two separate scenarios:
1. in which users may want to provide a persistence layer that fully restores job objects to a live state. To do so, if the executor does not guarantee that the job spec/info will be restored, the user must.
2. users can still set the wrong info after attach() leading to the same problem; so I don't think the issue is fully enforceable unless we somehow make certain properties to be write-once.

I'm writing this because I've kinda ran into such situations. But please consider this conversation resolved from my perspective (aka. press the "Resolve conversation" button unless you want to discuss further).

hategan (approved):

specification.md
andre-merzky: >     2\. users can still set the wrong info after attach()

I don't mind giving the user that freedom (and responsibility),  So, let's leave as it is (`MUST be NEW, attach MUST discard`), and reconsider once it turns out to be problematic (which I doubt).  Either way, thanks for the nitpick.

PR #77, opened 2020-12-22
fix #76

### LINKED ISSUE #76: naming consistency for specification classes
we use `JobSpecification` and `ResourcesSpec` - note the singular/plural difference, and the full name / short name.  I suggest to make this consistent with `JobSpecification` and `ResourceSpecification`.  

As argument for singular: It is customary to use singular in class names even when the can handle multiple entities, like `JobExecutor`, not `JobsExecutor`, `FileManager`, not `FilesManager`, etc.

andre-merzky:
Spec -> Specification
ResourcesSpec -> ResourceSpecification

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -24,7 +24,7 @@
 					- [Exceptions:](#exceptions)
 		- [Job](#job)
 			- [Methods](#methods)
-		- [JobSpecification](#jobspecification)
+		- [JobSpec](#jobspec)
 			- [Methods](#methods)
 		- [JobStatus](#jobstatus)
 			- [Methods](#methods)
@@ -38,9 +38,9 @@
 			- [Methods](#methods)
 		- [FaultDetail](#faultdetail)
 			- [Methods](#methods)
-		- [ResourceSpecification](#resourcesspecification)
+		- [ResourceSpec](#resourcespec)
 			- [Methods](#methods)
-		- [ResourceSpecificationV1](#resourcespecificationv1)
+		- [ResourceSpec](#resourcespecv1)
 			- [Methods](#methods)
 		- [JobAttributes](#jobattributes)
 			- [Methods](#methods)
@@ -49,7 +49,7 @@
 			- [Methods](#methods)
 		- [TimeUnit](#timeunit)
 	- [Appendices](#appendices)
-		- [Appendix A - Job Specification V1 Serialization Format](#appendix-a-job-specification-v1-serialization-format)
+		- [Appendix A - Job Spec V1 Serialization Format](#appendix-a-job-specification-v1-serialization-format)
 			- [Resources](#resources)
 				- [Reserved Resource Types](#reserved-resource-types)
 				- [V1-Specific Resource Graph Restrictions](#v1-specific-resource-graph-restrictions)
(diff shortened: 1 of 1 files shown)


hategan:
Now, hold on a second. We had the conversation about "Spec" vs. "Specification" and we settled for shortening it to "Spec".

I'd be happy to revisit this, but I don't think we should making it a habit of going back and forth on such things, unless there's a good reason.

andre-merzky:
> Now, hold on a second. We had the conversation about "Spec" vs. "Specification" and we settled for shortening it to "Spec".

Oh, I wasn't aware of that (or forgot that obviously), will change both to `Spec` then.

specification.md
```diff
@@ -49,7 +49,7 @@
 			- [Methods](#methods)
 		- [TimeUnit](#timeunit)
 	- [Appendices](#appendices)
-		- [Appendix A - Job Specification V1 Serialization Format](#appendix-a-job-specification-v1-serialization-format)
+		- [Appendix A - Job Spec V1 Serialization Format](#appendix-a-job-specification-v1-serialization-format)
```
SteVwonder, line 52: For consistency, maybe it makes sense to not have a space here:

```suggestion
		- [Appendix A - JobSpec V1 Serialization Format](#appendix-a-job-specification-v1-serialization-format)
```

specification.md
```diff
@@ -1080,10 +1080,10 @@ V1](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.ht
 which is itself a simplified version of the [canonical Flux jobspec
 format](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html).
 
-A jobspec V1 YAML document SHALL consist of a dictionary defining the
+A job specification V1 YAML document SHALL consist of a dictionary defining the
```
SteVwonder, line 1083: Were these changes from `jobspec` -> `job specification` still intended after the latest commit?

SteVwonder (commented):
Thanks @andre-merzky! Generally LGTM.  Mainly just a question if some `spec` -> `specification` changes were intended to still be included.

specification.md
andre-merzky: That's a good catch - I'll revert.

andre-merzky:
> Thanks @andre-merzky! Generally LGTM. Mainly just a question if some `spec` -> `specification` changes were intended to still be included.

Thanks for catching that - fixed.

specification.md
```diff
@@ -1083,7 +1083,7 @@ format](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1
 A jobspec V1 YAML document SHALL consist of a dictionary defining the
 resources, tasks and other attributes of a single program. The dictionary
 MUST contain the keys `resources`, `tasks`, `attributes`, and `version`.
-Each of the listed jobspec keys SHALL meet the form and requirements
+Each of the listed job specification keys SHALL meet the form and requirements
```
SteVwonder, line 1086: Sorry @andre-merzky . I should have been more verbose in my comment.  There are a hand full of these `jobspec` -> `job specification` changes that I think need to be reverted.   I won't tag them all to keep the noise on this PR down, but I count about 8 in total.

hategan:
Looks fine. Just trying to confirm one more time that "*Spec" is what we're going with (aka., if I remember **that** conversation correctly).

hategan:
Confirmed on slack that "Spec" is our goal

Issue #73, opened 2020-12-21
`waitFor` proposal


andre-merzky:
The [`waitFor`](https://github.com/ExaWorks/job-api-spec/blob/main/specification.md#job-waitfor) method currently accepts an timeout and a set of target states.  Following a [discussion with Stephen](https://github.com/ExaWorks/jpsi-python/pull/2/files#r546202828), I suggest the following changes

  - rename `waitFor` to `wait` (in sync with other job and process APIs)
  - let the call always return when a final state is reached.

About that last item: assume `job.wait(timeout=None, states=[ACTIVE])`, and the job fails to get queued, so it goes to `FAILED` before ever passing `ACTIVE`.  The call would now hang forever.  In general, waiting on jobs in a final state is rather useless.  We could also consider raising an exception if the job goes into a final state which is not explicitly listed as a target state though.

hategan:
> The [`waitFor`](https://github.com/ExaWorks/job-api-spec/blob/main/specification.md#job-waitfor) method currently accepts an timeout and a set of target states. Following a [discussion with Stephen](https://github.com/ExaWorks/jpsi-python/pull/2/files#r546202828), I suggest the following changes
> 
> * rename `waitFor` to `wait` (in sync with other job and process APIs)

wait() is a final method in Java, so, in Java it can't be wait().

> * let the call always return when a final state is reached.

That makes sense. The purpose of waitFor() is to wait until a condition is reached. There is no point in waiting for something that will never happen.

> 
> About that last item: assume `job.wait(timeout=None, states=[ACTIVE])`, and the job fails to get queued, so it goes to `FAILED` before ever passing `ACTIVE`. The call would now hang forever. In general, waiting on jobs in a final state is rather useless. We could also consider raising an exception if the job goes into a final state which is not explicitly listed as a target state though.

Yes, I think it might make sense to not follow the normal flow when abnormal things happen. So an exception might make sense.

andre-merzky:
> wait() is a final method in Java, so, in Java it can't be wait().

I understand then that the Java binding will have `waitFor`, but that should not affect all other languages, right?

hategan:
> > wait() is a final method in Java, so, in Java it can't be wait().
> 
> I understand then that the Java binding will have `waitFor`, but that should not affect all other languages, right?

It was my impression that we're aiming for a spec that doesn't leave that much room for interpretation.

And I also don't think we should spend time on naming issues unless the offence is particularly grave. That said, whatever...

[...]

PR #79, opened 2020-12-22
add FINAL enumeration

andre-merzky:
issue #73 (job.waitFor(states = [final_states])) reminds me of something: we do define state enums.  I find it very useful to also define a constant `FINAL` which refers to the set of final states.  That allows constructs like:
```py
if job.state not in jpsi.FINAL:
   job.wait(states=jpsi.FINAL + [jpsi.ACTIVE])
```
etc.  Of course, that is only syntactic sugar as al this can be expressed as
```py
if job.state not in [jpsi.DONE, jpsi.FAILED, jpsi.CANCELED]:
   job.wait(states=[jpsi.DONE, jpsi.FAILED, jpsi.CANCELED, jpsi.ACTIVE])
```
but `FINAL` does make this a fair bit easier to handle for the end user...

This PR adds that `FINAL` constant as additional state enum.


hategan:
Can we discuss such things instead of going straight to a PR?

1. Where would this constant live? There are no package constants (e.g. `jpsi.ACTIVE`) in the spec.
2. `wait(FINAL + [...])` is superfluous if we accept the part of #73 that has `wait()` always return when a final state is reached.
3. There is a perfectly reasonable solution to `job.state.status not in FINAL`: `job.state.isFinal()`

Basically, with #73 and what's already there, the equivalent of your code is:
```python
if not job.state.final:
   job.wait(jpsi.ACTIVE)
```

I'd argue that's more concise, not to mention more efficient.

andre-merzky:
> Can we discuss such things instead of going straight to a PR?

See slack...


>     1. Where would this constant live? There are no package constants (e.g. `jpsi.ACTIVE`) in the spec.

In the same place where the state enumerations live

 
>     2. `wait(FINAL + [...])` is superfluous if we accept the part of #73 that has `wait()` always return when a final state is reached.

As per this [comment](https://github.com/ExaWorks/job-api-spec/issues/73#issuecomment-749157189) (and other comments in that thread), the `wait` call will raise an exception when not explicitly waiting for final transitions, so one would need to **always** include the final states in the arguments, which is precisely what this PR addresses.  


>     3. There is a perfectly reasonable solution to `job.state.status not in FINAL`: `job.state.isFinal()`
> 
> Basically, with #73 and what's already there, the equivalent of your code is:
> 
> ```python
> if not job.state.final:
>    job.wait(jpsi.ACTIVE)
> ```
> I'd argue that's more concise, not to mention more efficient.

Alas, per the exception above, that code turns into either:
```py
if not job.state.final:
    job.wait([jpsi.ACTIVE, jpsi.COMPLETED, jpsi,CANCELED, jpsi.FAILED])
    if job.state == jpsi.ACTIVE:
        # do stuff
```

or

```py
if not job.state.final:
    try:
        job.wait(jpsi.ACTIVE)
        # do stuff
    except:
        pass
```

hategan:
> In the same place where the state enumerations live.

If this is an actual enum, strictly typed languages might have an issue with this.

> ```python
> if not job.state.final:
>     try:
>         job.wait(jpsi.ACTIVE)
>         # do stuff
>     except:
>         pass
> ```

If we go the exception path, yes. If we have `wait()` return the state that triggered the wait to end, no. Either way, there is no need to provide multiple ways of doing one thing.

I could also note that the *if* test is unnecessary if `wait()` exits (with or without exception) if the job cannot reach the desired state. So you can shorten things to:

```python
job.wait(JobState.ACTIVE) # if wait() does not throw exception on final states

with suppress(UnreachableStateException):
    job.wait(JobState.ACTIVE)
```

hategan:
> In the same place where the state enumerations live.

If this is an actual enum, strictly typed languages might have an issue with it.

> ```python
> if not job.state.final:
>     try:
>         job.wait(jpsi.ACTIVE)
>         # do stuff
>     except:
>         pass
> ```

If we go the exception path, yes. If we have `wait()` return the state that triggered the wait to end, no. Either way, there is no need to provide multiple ways of doing one thing.

I could also note that the *if* test is unnecessary if `wait()` exits (with or without exception) if the job cannot reach the desired state. Not only that, it is a race condition in the making. The job may very well transition to a final test after the `if` but before `wait()`, so if you want to write correct code, it needs to handle the output/exception from `wait()`:

```python
if job.wait(JobState.ACTIVE) == JobState.ACTIVE: # if wait() does not throw exception on final states
    doStuff()
--- or ---
try:
    wait(JobState.ACTIVE)
    doStuff()
except UnreachableStateException:
    pass
```

andre-merzky:
>>  In the same place where the state enumerations live.

> If this is an actual enum, strictly typed languages might have an issue with it.

We'll figure that out in the language binding then.


> If we go the exception path, yes.

Ack, this PR is prompted by that.  I read in #73 though that we do have agreement on the exception path - do you read that differently?

hategan:
Ok. I'm voting NO on this:

- the use case contains a race condition; addressing that race condition invalidates the use case
- with #73 adopted such that either `wait()` returns the reached state or throws an exception, there already exists a way of doing what this PR is trying to address
- making it "a fair bit easier" is not a good enough reason to add features
- the proposed solution has questionable bindings to strictly-typed languages that are not addressed ("we'll do it later/somewhere else" is not acceptable)

andre-merzky:
> Ok. I'm voting NO on this:
> 
>     * the use case contains a race condition; addressing that race condition invalidates the use case

The example is too artificial (it tries to show two uses of `FINAL` in one example) - but also it is not much of a race, the wait would immediately return.


>     * with #73 adopted such that either `wait()` returns the reached state or throws an exception, there already exists a way of doing what this PR is trying to address

I would like to ask again if you read #73 the same way as me, that it adopts the exception raising option?  But to the point:  
that is like saying we don't need `is_final` because one can do `if state is not in [COMPLETED, FAILED, CANCELED]`.  The statement is true, but somewhat misses the point.  Yes, we can do without, but the API is more useful and usage is more concise when having it.

 
>     * making it "a fair bit easier" is not a good enough reason to add features

What would be?

 
>     * the proposed solution has questionable bindings to strictly-typed languages that are not addressed ("we'll do it later/somewhere else" is not acceptable)

Ok, I bite.  I am really not a java programmer, so had to google around:
```java
import java.util.*;

class States{

    public static int NEW       = 0;
    public static int QUEUED    = 1;
    public static int ACTIVE    = 2;
    public static int COMPLETED = 3;
    public static int FAILED    = 4;
    public static int CANCELED  = 5;

    public static final List FINAL = List.of(COMPLETED, FAILED, CANCELED);
}

class Test{
    public static void main(String args[]){
        int t = States.FAILED;
        if (States.FINAL.contains(t)) {
            System.out.printf("Hello JAVA %d (final)\n", t);
        } else {
            System.out.printf("Hello JAVA %d\n", t);
        }
    }
}
```

This is probably not quite right for reasons I don't know about - but really, can you see no way to implement something like that in Java, at all?  Is it really impossible to define a list of constants somewhere which can be used for this?  

But even if: we *do not define the Java binding here*.  If Java can't do that, then the Java rendering won't have that.  Just like the C rendering won't have exceptions, or a shell rendering would have not callbacks (Making this up, but anyway).

PS: I am frustrated (as I am sure you are) that we seem to be happy to bang our heads together on seemingly every single issue these days.   It's quite fascinating also, in a way ...   Either way, sorry for that, really, I owe you another beverage of your choice...

hategan:
Sorry Andre, but this is crossing into arguments territory. I'd rather not pursue that avenue.

I cannot stop you from merging this PR. But I do not agree with it.

Re #73, I read it as saying "hey, exception looks like a good idea". However, I believe #73 needs more work. I'll add my comments there.

andre-merzky:
> Sorry Andre, but this is crossing into arguments territory. I'd rather not pursue that avenue.

Hmm, how are we going to discuss without arguments?  Or am I parsing this incorrectly, maybe?  I don't like the long threads either, but are the arguments wrong?

> I cannot stop you from merging this PR. But I do not agree with it.

Yes, you can.  Per our agreement: the last one (not the author) resolving a discussion merges the PR.  So unless you mark this as resolved, nothing will happen.
 
> Re #73, I read it as saying "hey, exception looks like a good idea". However, I believe #73 needs more work. I'll add my comments there.

Ok, thanks - that might change the premise of this PR.

hategan:
> > Sorry Andre, but this is crossing into arguments territory. I'd rather not pursue that avenue.
> 
> Hmm, how are we going to discuss without arguments? Or am I parsing this incorrectly, maybe? I don't like the long threads either, but are the arguments wrong?

Not wrong (except the race), but some I find to be a bit stretched. Let me know if you want me to be specific.

> 
> > I cannot stop you from merging this PR. But I do not agree with it.
> 
> Yes, you can. Per our agreement: the last one (not the author) resolving a discussion merges the PR. So unless you mark this as resolved, nothing will happen.

It is an amicable agreement, not a UNIVERSE WILL COLLAPSE ON ITSELF IF YOU DO IT sort of deal. You could also say "I hear you, but I want to merge it anyway" and I'll approve it. Or Stephen can independently approve it without "breaking" any "rules".

> 
> > Re #73, I read it as saying "hey, exception looks like a good idea". However, I believe #73 needs more work. I'll add my comments there.
> 
> Ok, thanks - that might change the premise of this PR.

As far as the use case above, yes. But then, again, if the user can implement this in one line, maybe we shouldn't do it.

> Either way, sorry for that, really, I owe you another beverage of your choice...

Same. Fair warning though: we might end up in a coma!

I mean, discussion is inevitable and desirable in many cases. What I worry is that we won't finish much if we discuss too much. So I'd rater have a reasonable solution now and choosing between roughly equivalent choices is left for another time.

hategan:
Given the recent modifications to `Job.wait()`, is further discussion here still useful?

andre-merzky:
> Given the recent modifications to `Job.wait()`, is further discussion here still useful?

I find `FINAL` still useful, but given the changes to `wait()` not urgent anymore, so don't mind closing for now.

PR #81, opened 2020-12-25
address issue #41 (`attach()`), no `list()`

### LINKED ISSUE #41: Accomodating short-lived clients

andre-merzky:
Same as #68, but no list, as that needs more discussion in #69.

Closes #41


hategan:
Sorry to nitpick.

- If we go with both JPSI ID and native ID, then attach should use the native ID IMO. There would be no way to attach using the JPSI id without storing some mapping between the two or guaranteeing that the native ID is somehow contained in the JPSI id.
- Are we not planning to address the race in notifications when attaching? (i.e. have `attach(job, nativeId)` such that callbacks can be added) at this time?

andre-merzky:
> Sorry to nitpick.

Not a problem :-)

>  * If we go with both JPSI ID and native ID, then attach should use the native ID IMO. There would be no way to attach using the JPSI id without storing some mapping between the two or guaranteeing that the native ID is somehow contained in the JPSI id.

Yeah, probably.  Lets wait for that discussion to settle.  I know you are not in favor of `list()`, but do you have an opinion on the `ID` discussion in #68?


>     * Are we not planning to address the race in notifications when attaching? (i.e. have `attach(job, nativeId)` such that callbacks can be added) at this time?

I am personally not too worried about that, really: one can miss state transitions before the `attach` is successful anyway, and I can't think of a use case that would need the cb to be registered early.  That doesn't mean that those use cases don't exist, and I don't oppose addressing the issue.

In #41, you wrote:
> We need to think about this a bit, since returning a job is probably a bad idea if you want to add a callback. It would lead to possible race conditions. So maybe JobExecutor.attach(Job job, String nativeId) would be a better choice, such that client code can add the callback before invoking attach().

That would be one way to address it - but it needs clarification on possible call sequences:

```py
job = jpsi.Job()
spec = jpsi.JobSpecification()
spec.executable = '/bin/sleep'
spec.arguments = ['10']
job.specification = spec
je.submit(job)
je.attach(job)
```

We need to specify that `attach` expects jobs to be in `NEW` state, have no `jobspec` attached, etc.

Another option would be :
```py
job = je.attach(job_id='foo', cb=bar)
```

which would imply that the given callback is attached before the job is returned.  I don't like that too much either really: it's not very intuitive (to me), and opens a race to the callback being invoked before the job is assigned in the client code.

I'm happy to follow other people's opinion on this one.

hategan:
> > * If we go with both JPSI ID and native ID, then attach should use the native ID IMO. There would be no way to attach using the JPSI id without storing some mapping between the two or guaranteeing that the native ID is somehow contained in the JPSI id.
> 
> Yeah, probably. Lets wait for that discussion to settle. I know you are not in favor of `list()`, but do you have an opinion on the `ID` discussion in #68?

Did anybody say "opinion"? I have those!
Sorry, I didn't get to say much there yet. I'll try to get to it soonish.

> 
> > ```
> > * Are we not planning to address the race in notifications when attaching? (i.e. have `attach(job, nativeId)` such that callbacks can be added) at this time?
> > ```
> 
> I am personally not too worried about that, really: one can miss state transitions before the `attach` is successful anyway, and I can't think of a use case that would need the cb to be registered early.

The flow there is that you'd attach to a completed job, `attach()` would complete very quickly and send its events. By the time you get to call `setStatusCallback()` on the job, you've already missed that state. This isn't the same as missing states that happened before `attach()`. There is some assumption that `attach()` will figure out what has happened to the job even thought that might have happened before `attach()`.

So I think this is quite important.

> That doesn't mean that those use cases don't exist, and I don't oppose addressing the issue.
> 
> In #41, you wrote:
> 
> > We need to think about this a bit, since returning a job is probably a bad idea if you want to add a callback. It would lead to possible race conditions. So maybe JobExecutor.attach(Job job, String nativeId) would be a better choice, such that client code can add the callback before invoking attach().
> 
> That would be one way to address it - but it needs clarification on possible call sequences:
> 
> ```python
> job = jpsi.Job()
> spec = jpsi.JobSpecification()
> spec.executable = '/bin/sleep'
> spec.arguments = ['10']
> job.specification = spec
> je.submit(job)
> je.attach(job)
> ```
> 
> We need to specify that `attach` expects jobs to be in `NEW` state, have no `jobspec` attached, etc.

Technically, it would be `je.attach(job, nativeId)`. Since you cannot set the nativeId on a job, `attach()` could check if either the job's nativeId is empty, the same as the nativeId passed to `attach()`, or a different one. The first two cases should not trigger an exception. That would nicely allow the above example to work just fine.

> 
> Another option would be :
> 
> ```python
> job = je.attach(job_id='foo', cb=bar)
> ```
> 
> which would imply that the given callback is attached before the job is returned. I don't like that too much either really: it's not very intuitive (to me), and opens a race to the callback being invoked before the job is assigned in the client code.

Yeah, I don't like it that much either.

> 
> I'm happy to follow other people's opinion on this one.

hategan:
I think we agreed on a call that attaching with something else than the native ID is not ideal at this time.

Andre, if you agree on the need to address the race condition mentioned above, can you update this to `JobExecutor.attach(job, nativeId)`?

SteVwonder:
From the ☎️ call:

calling `attach` on a native job id and J/PSI job object should not only enable state callbacks but also populate the J/PSI job object with state and specification information.  That information should be obtained asynchronously by the implementation (for performance reasons) and the first state event callback for the job should **not** fire until the complete job information is populated (so that the user doesn't not have to worry about a partially intialized/populated J/PSI job object).

Issue #72, opened 2020-12-18
How to initialize a J/PSI Connector


SteVwonder:
There was a discussion of how to initialize a particular J/PSI connector. 

Should the connector be stateful or stateless?  What are the pros and cons of each?  Once we determine that, then we should decide what the initialization should look like (e.g., factory, class constructor, etc).

This is the main issue blocking the prototype in https://github.com/ExaWorks/jpsi-python/pull/2 from being standards-compliant.

hategan:
I tried to write an "abstract" version of what follows, but I think it ends up more being unnecessarily obscure. So I'll try simple sentences and not that much abstraction.

We need to do two things, which, from a certain distance, can be seen as one:
A. Instantiate a connector
B. Configure a connector

There are probably more ways to do this, but here are a few that I can think of:
1. JobExecutor is a base class and we use constructors to instantiate connectors, which are subclasses of JobExecutor:
    - `e1 = PBSJobExecutor(); e2 = SLURMJobExecutor()`
    - there are some possible ways to do (B):
        1. By passing configuration parameters to a constructor:
            - `e3 = FluxJobExecutor(endpoint1); e4 = FluxJobExecutor(endpoint2)`
        2. By using an initialization method:
            - `e5 = FluxJobExecutor(); e5.initialize({endpoint: endpoint1}); e6 = FluxJobExecutor(); e6.initialize({endpoint: endpoint2})`
        3. As part of the job specification:
           - `e7 = FluxJobExecutor(); j1.setEndpoint(endpoint1); j2.setEndpoint(endpoint2); e7.submit(j1); e7.submit(j2);`
2. JobExecutor is a base class, but we use a factory to instantiate connectors; this can coexist with (1):
    - `e8 = JobExecutorFactory.getInstance("PBS"); e9 = JobExecutorFactory.getInstance("SLURM")`
    - (B):
        1. Pass configuration parameters to the factory:
            - `e10 = JobExecutorFactory.getInstance("Flux", endpoint1); e11 = JobExecutorFactory.getInstance("Flux", endpoint2);`
        2. Using an initialization method once we get a (new) instance through the factory
        3. As part of the job spec, as above
3. JobExecutor (or something that implements the JobExecutor interface) is a frontend and it is its own business what it does behind the scenes:
    - `e12 = JobExecutor("PBS"); e13 = JobExecutor("SLURM")`
    - factories make little sense here
[...]

PR #85, opened 2021-01-08
Add class descriptions

### LINKED ISSUE #4: Add brief description for each class

hategan:
Fixes #4

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -942,6 +942,9 @@ signifies to the LRM that GPU nodes are being requested.
 
 ### JobAttributes
 
+A class containing ancillary job information that describes how a job is to be
+run.
+
 #### Methods
 
 <a name="jobattributes-setduration"></a>
--- a/specification.md
+++ b/specification.md
@@ -259,6 +259,39 @@ customary in the language in which the library is implemented
 
 ### JobExecutor
 
+The `JobExecutor` represents one or more concrete mechanisms for
+executing jobs. It contains all the operations that are specific to a
+particular such mechanism. Specifically, it knows how to start a job
+through the `submit()` call, query the status of jobs and inform client
+API of any updates through callbacks, and can `cancel()` a running job.
+
+Client code interacts with a concrete job execution mechanism by invoking
+methods on objects declared (in a strictly-typed language) as
+`JobExecutor`. This leaves a number of possible ways to structure an
+implementation of  this API:
+
+1. Treat `JobExecutor` as an abstract base class and have concrete
+subclasses of `JobExecutor` implement the specific mechanisms. The
+subclasses can then be instantiated either directly, using a factory
+pattern, or any other reasonable mechanism. For example:
+
+    ```java
+    JobExecutor executor = new PBSJobExecutor();
+	Job job = ...
+	executor.submit(job);
+	```
+
+2. Treat `JobExecutor` as a frontend class, which can possibly be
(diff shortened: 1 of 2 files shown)


specification.md
```diff
@@ -259,6 +259,39 @@ customary in the language in which the library is implemented
...
+The `JobExecutor` represents one or more concrete mechanisms for
+executing jobs. It contains all the operations that are specific to a
+particular such mechanism. Specifically, it knows how to start a job
+through the `submit()` call, query the status of jobs and inform client
+API of any updates through callbacks, and can `cancel()` a running job.
+
+Client code interacts with a concrete job execution mechanism by invoking
+methods on objects declared (in a strictly-typed language) as
+`JobExecutor`. This leaves a number of possible ways to structure an
+implementation of  this API:
```
SteVwonder, line 271: Is the intention behind exposing multiple ways to structure this class to support different languages, with each language only using one possible way?  Or is the intention to have both ways supported within the same implementation?

My main reason for asking is out of source-code compatibility concerns.

SteVwonder (commented):
Generally LGTM.  Just one question below:

specification.md
hategan: I think it's because, as far as I remember, we didn't reach a consensus on one or the other. Or, rather said, each choice may be more appropriate for a specific scenario, such as implementing the API from scratch or writing a wrapper for an existing library.

I wrote a bit on this in #72, but there are other places where discussions around the exact nature of `JobExecutor` happened.

specification.md
```diff
@@ -348,6 +381,20 @@ to `null`.
...
+as well as the job's state. Instances of this class are created by user
+code and populated with information describing what to run as part of the
+job (e.g., the executable path, the arguments, etc.) as well as how the
+job is to be run, where applicable. The later involves specifying, for
+example, the number of CPU cores desired or other such requirements. Once
+all relevant information is provided, the job may be sent to an
+underlying implementation using the [`submit()`](#jobexecutor-submit)
+call of a [`JobExecutor`](#jobexecutor) instance. The executor then takes
+care of updating the status of the job, which is accessible synchronously
+through the [`Job.getStatus()`](#job-getstatus) call or. asynchronously,
```
andre-merzky, line 394: ` or. ` should be ` or, `

andre-merzky (approved):
LGTM.  I echo @SteVwonder's point about source code portability.

specification.md
hategan: Thanks for catching that.

hategan:
> LGTM. I echo @SteVwonder's point about source code portability.

So do I. I can either remove that part or we can clarify once we reach a consensus.

specification.md
andre-merzky: >>    LGTM. I echo @SteVwonder's point about source code portability.

> So do I. I can either remove that part or we can clarify once we reach a consensus.

I would suggest to add a statement along the lines of:

```
Language bindings SHOULD explicitly define an mechanism to ensure source code compatibility between implementations.
```

hategan: I added a note to this extent. Please let me know if that works. If not I can also replace with the above suggested statement.

andre-merzky: Works for me, thanks

SteVwonder: Works for me as well. Thanks!

SteVwonder (approved):
LGTM! Thanks @hategan.

PR #87, opened 2021-01-11
Added SubmitException to the doc.

hategan:
Although, I do wonder: should it perhaps be called `SubmissionException`?


andre-merzky (changes requested):
`SubmitException` sounds just right to me.  I would suggest to keep the same methods as for the `InvalidJobException`, i.e., to also include the `getJob()` method.  Otherwise this looks good to me.

hategan:
> lso include the `getJob()` method.

Yeah, I was thinking about that. But SubmitException is, by design, not associated with a particular job except by the submit() call. Anyway, long story short, I wasn't sure but was leaning towards not having it.

Can you think about it a bit more and see if that changes your mind. If not, I can add getJob().

SteVwonder:
Is the intent that this would be thrown (for example) when a connection to the underlying RM fails or when the user hits a center policy limit on the max number of simultaneous jobs?

hategan:
> Is the intent that this would be thrown (for example) when a connection to the underlying RM fails or when the user hits a center policy limit on the max number of simultaneous jobs?

We should clarify this. But I see it as the former. The latter would be "InvalidJobException". But now that you mention it, I'm having doubts.

I do believe that there should be an exception that reflects technical issues with the connector, the connector-LRM interface, and the LRM. I thought SubmitException was it. But then the scenario you describe it not really a problem with the job specification, in that it does not appear to directly fit "fails for a reason that is independent of the job being submitted", which is InvalidJobException.

Thoughts?

andre-merzky:
> > Also include the `getJob()` method.
> 
> Yeah, I was thinking about that. But SubmitException is, by design, not associated with a particular job except by the submit() call. Anyway, long story short, I wasn't sure but was leaning towards not having it.
> 
> Can you think about it a bit more and see if that changes your mind. If not, I can add getJob().


I honestly don't see a reason for `getJob` on the `InvalidJobException` either, since the job object should be available when the `submit` call raises that exception.  I think the `getJob` mostly makes sense on bulk submissions to easily correlate exception and jobs - but we don't have bulks right now.  My argument was mostly for symmetry: both exceptions are raised on the `submit` call, both related to the action of  submitting one specific job - I don't see why one should have the job attached but not the other...

> `InvalidJobException` vs. `SubmitException`

I think the problem is that the client side library is ultimately not the authorized entity to decide when a job is valid or not: that is the batch system.  The library can decide if the bast data are valid types and values (any thus could raise `ValueError` (requesting `-2` cores) or `TypeError` (requesting `1.5` cores) - but as long as the description passes those types and value checks, the job should be submittable.  The submission to the backend is not necessarily synchronous (that is why we have the separate `QUEUED` state, which is reached after backend submission), so the decision point for `InvalidJob` may in many cases not even make it into a `submit` exception, but is more likely to move the job into a `FAILED` state.

In that sense, I considered the `InvalidJobException` as a combination of type and value error until now, and not related to the validity of the job from the perspective of the batch system.

hategan:
> I honestly don't see a reason for `getJob` on the `InvalidJobException` either, since the job object should be available when the `submit` call raises that exception.

Right. It's meant as `getX()` when you have `InvalidXException`, not much more.

> I don't see why one should have the job attached but not the other...
> 

The same reason why a hypothetical `open()` could throw a `FileNotFoundException` that has a `getFile()`, but also an `IOError` like `catastrophic disk failure` for which `getFile()` would make less sense.

That said, this is a thing and I'm fine removing `getJob()`.


>> `InvalidJobException` vs. `SubmitException`
>
> [... and the issue of API-side vs. LRM side validation...]

The API doesn't exactly define how `submit()` should do its job. It could simply put the job in a queue and have another thread send jobs from that queue to the LRM. Or, it could write a job script and call `qsub`. If the latter, it could interpret any potential errors from the LRM and decide if they are a job exception or something else. In the former case (`submit()` queues jobs), the `InvalidJobException` would presumably be attached to the failed status. So how `submit()` works doesn't quite matter when it comes to this exception.

There is a note about this attached to `JobExecutor.submit()`.

hategan:
I removed `getJob()` from `InvalidJobException` based on:

> I honestly don't see a reason for `getJob` on the `InvalidJobException` either

Let me know if this works better or whether we should have both `InvalidJobException` and `SubmitException` have a `getJob()`.

andre-merzky:
> Let me know if this works better or whether we should have both `InvalidJobException` and `SubmitException` have a `getJob()`.

This works for me - thanks!

> I do believe that there should be an exception that reflects technical issues with the connector, the connector-LRM interface, and the LRM. I thought SubmitException was it. But then the scenario you describe it not really a problem with the job specification, in that it does not appear to directly fit "fails for a reason that is independent of the job being submitted", which is InvalidJobException.
>
> Thoughts?

I see several options, and I am fine with either one
  - cover connection issues or other implementation related problems under `SubmitException`, 
  - introduce a new exception for all failures which are not related to the batch system refusing the job.
  - relying on custom language exceptions to convey those problems 

I would like to note that our exception list is incomplete anyway, to some extent, as we don't completely specify exceptions on other methods either.  `wait()` for example might raise if the internal job information is purged, `setStatusCallback` might raise if the parameter is not a callable type, etc.  I think that potential implementation level exceptions for `submit()` fall into the same category.  I would not mind if we maintain the current wording in that respect and leave more details to language bindings or later iterations of this document.

andre-merzky:
PS.: for me this is good to merge - @SteVwonder ?

hategan:
> `etStatusCallback` might raise if the parameter is not a callable type

That would be a compiler error in a strict language (or a mypy error, if used, in python). So I'm mostly concerned with what Java would call "checked exceptions" (i.e., those that make it into a method's signature and which must be handled by the caller or explicitly added to the caller's signature), whereas other things like failure to allocate memory would be runtime exceptions.

andre-merzky:
> checked exceptions  ... runtime exceptions

Yeah, I am with you there...

specification.md
```diff
@@ -897,13 +897,40 @@ debugging purposes, but which should not, in general, be presented to an
 end-user.
 
 
-<a name="invalidjobexception-getjob"></a>
+
+### SubmitException
+
+This exception is thrown when the
+[`JobExecutor.submit()`](#jobexecutor-submit) call fails for a reason
+that is independent of the job being submitted.
```
SteVwonder, line 905: Sorry. Really minor nit but this was confusing to me on the second read through.  As written, it kind of sounds like "call fails for a reason that is independent of the job submission".  My suggestion:

```suggestion
that is independent of the job that is being submitted.
```

specification.md
```diff
@@ -897,13 +897,40 @@ debugging purposes, but which should not, in general, be presented to an
...
+
+#### Methods
+
+<a name="submitexception-getdetail"></a>
 ```java
-Job getJob()
+FaultDetail getDetail()
```
SteVwonder, line 911: Do we have an issue open about detailing `FaultDetail`?  I see it is also used for `InvalidJobException`.  So adding it here is fine with me, but I just want to make sure we have that as a recorded TODO.

SteVwonder (approved):
LGTM!  Two non-blocker comments below:

specification.md
SteVwonder: PS - a search for `faultdetail` returns nothing in our issue tracker

SteVwonder:
Thanks @hategan! Double approval now, so feel free to merge once you are happy with it/done pushing changes.

PR #88, opened 2021-01-12
The idea of having the "Impelementation note" tag as vertical text on…

hategan:
… the left didn't work well with short notes, so this got moved to the top instead.


andre-merzky (approved):
Thanks

PR #89, opened 2021-01-13
Added text (and a diagram) in the beginning of Layer 0 that gives an …

hategan:
…overview of the components.

One thing that occurred to me is that we're somewhat conflating resource requirements with the job description. For example, "number of nodes" is a resource requirement, but "number of processes" is not. We should probably fix this.


specification.md
```diff
@@ -231,11 +231,41 @@ implementations is detailed in [`Job.getStatus()`](#job-getstatus) and
 
 ## The Job API; Layer 0
 
-This section describes the basic local API.
-
-
-
-###Implementation Notes
+This section describes the most basic form of J/PSI API, namely one in
+which the location of the backend is either implicit or local. The main
```
andre-merzky, line 235: > `implicit or local` 

Maybe change to `implicit and local`?  As long as we don't have an parametrized c'tor, the backend location will have to remain implicit.

andre-merzky (approved):
Apart from a minor nitpick, this looks good to me - thanks!

specification.md
hategan: That would leave out Flux and I didn't want to leave out Flux.

I think we can still ponder on the topic of how to pass contact information to the executor in a way that doesn't involve the constructor so that we can have Flux be part of Layer 0.

I'm also thinking that there's two Layer 1s. One in which we stick a custom client/service infrastructure on top of Layer 0 and another in which we abstract over natively remote APIs.

andre-merzky: FWIW, Flux works implicitly: w/o explicit contact URL, it will pick the contact endpoint from the process environment.  It holds up a RADICAL-SAGA backend - that needs an explicit URL.  But the comment was a nitpick anyway and should not hold up the PR.

PR #92, opened 2021-01-16
Separate some implementation notes; moved two of those to Layer 0 whe…

hategan:
…re they seem to better belong

This basically adds the nice yellow background to a few paragraphs.


andre-merzky (approved):
LGTM, thanks.

Issue #52, opened 2020-11-20
Add more examples of API use


hategan:
Some examples have been added by PR #47. However, not all requested examples in the initial TODO are added by that PR. The reasons are listed in the PR comments. This is an issue mean to track the remaining examples in the TODO item so that the PR can be merged as is for now.

SteVwonder:
For the "exotic"/"complex" requests, I think it would be good to have a few examples that construct some "complex" ResourceSpecifications (doesn't need to be the full job). Two in particular:

- P total processes where each process gets C cpus and G gpus
- N exclusive nodes, each with P processes

It would also be good to have an example that constructs a JobAttribute, sets a value on it, and then adds that JobAttribute to a JobSpecification.

An example that changes the environment would also be a good one.

hategan:
I'll add a couple more items here in an effort to move away from an in-document todo list. They are:

- [ ] Submit a malformed or unsatisfiable job, then check for the error and print it out
- [ ] Construct a job that uses all the various “knobs” of the resource and job specifications (with some verbose comments thrown in)

hategan:
I am merging this issue into #18.

PR #94, opened 2021-01-19
add additional examples

andre-merzky:
I think this covers the examples listed in #52.  Happy to iterate on them and/or to add additional examples, but would like to get a quick feedback.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1553,7 +1553,7 @@ optimize the throughput of connections to particular services.
 
 
 
-### Appendix C - examples
+### Appendix C - Examples
 
 This Appendix contains some examples of how the API can be used. Unlike the
 specification language, the examples are in a hypothetical Python binding, which
@@ -1584,7 +1584,7 @@ for i in range(N):
     jex.submit(job)
 
 for i in range(N):
-    jobs[i].wait_for()
+    jobs[i].wait()
 ```
 
 
@@ -1603,22 +1603,22 @@ class ThrottledSubmitter:
         self.jex = jpsi.JobExecutorFactory.get_instance('torque', '>= 0.2')
         # keep track of completed jobs so that we can submit the rest
         self.jex.set_status_callback(self.callback)
-        self.crt = 0
+        self.count = 0
 
     def make_job(self):
         ...
 
     def submit_next():
-        if self.crt < N:
-            jex.submit(self.jobs[self.crt])
-            self.crt += 1
+        if self.count < N:
+            jex.submit(self.jobs[self.count])
+            self.count += 1
 
     def start(self)
         # create list of jobs
         self.jobs = [self.make_job() for i in range(N)]
 
         # submit initial M jobs
-        while self.crt < M:
+        while self.count < M:
             submit_next()
 
     def callback(self, job, status):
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -1631,6 +1631,168 @@ ThrottleSubmitter().start()
...
+spec_2.executable = True   # type error
+
+job_1.specification = spec_1
+job_2.specification = spec_2
+
+try:
+    jex.submit(job_1)
+except jpsi.InvalidJobException:
+    # this should not happen
+    assert(True)
```
SteVwonder, line 1658: Should this `assert(False)` to throw an assertion error in the exceptional case that this does occur?

specification.md
```diff
@@ -1631,6 +1631,168 @@ ThrottleSubmitter().start()
...
+except jpsi.SubmitException:
+    # this *can* happen, dependent on backend state and policies
+    print('could not submit job - try again later')
+
+try:
+    jex.submit(job_2)
+except jpsi.InvalidJobException as e:
+    print('submission failed: %s' % e)
+else:
+    assert(True)  # the above should have raised an `InvalidJobException`
```
SteVwonder, line 1668: Same question as above

specification.md
```diff
@@ -1631,6 +1631,168 @@ ThrottleSubmitter().start()
...
+job.specification = spec
+
+jex.submit(job)
+job.wait()
+```
+
+
+#### N exclusive nodes, each with P processes
+
+This example will place a job across 5 nodes with 3 ranks per node.
```
SteVwonder, line 1710:
```suggestion
This example will place a job across 5 nodes with 2 ranks per node.
```

specification.md
```diff
@@ -1631,6 +1631,168 @@ ThrottleSubmitter().start()
...
+C = 4
+G = 1
+
+res_spec = jpsi.ResourceSpecification({
+              'process_count'    : P,
+              'cores_per_process': C,
+              'gpus_per_process' : G,
+           })
+job_spec = jpsi.JobSpecification({
+              'executable': 'workload.py',
```
SteVwonder, line 1692:
```suggestion
              'executable': 'echo',
```

Assuming `workload.py` isn't provided somewhere else, this example will error out if someone blindly copies and pastes it.  Maybe we want to sacrifice some "realism" to increase the probability that it "just works"?  Just a though.  Feel free to ignore.

specification.md
```diff
@@ -1631,6 +1631,168 @@ ThrottleSubmitter().start()
...
+              # other jobs can run on unused resources of the job's nodes
+              'exclusive_nodes': false,
+              # run a total of 10 ranks
+              'process_count': 10,
+              # place 3 ranks per node (last rank will be placed alone)
+              'processes__per_node': 3,
+              # each rank obtains 4 cores
+              'cores_per_process': 4,
+              # … and 2 GPUs
+              'gpus_per_process': 4,
```
SteVwonder, line 1757:
```suggestion
              'gpus_per_process': 2,
```

SteVwonder (commented):
Awesome! I think these are great. Just a few nits below:

specification.md
andre-merzky: of course!

andre-merzky:
Thanks @SteVwonder!

specification.md
andre-merzky: What was I thinking?    Nothing apparently :-P  Thanks.

specification.md
```diff
@@ -1631,6 +1631,168 @@ ThrottleSubmitter().start()
...
+job_2.specification = spec_2
+
+try:
+    jex.submit(job_1)
+except jpsi.InvalidJobException:
+    # this should not happen
+    assert(False)
+except jpsi.SubmitException:
+    # this *can* happen, dependent on backend state and policies
+    print('could not submit job - try again later')
```
hategan, line 1661: This would be a good place to use the `message` field of `SubmitException`.

specification.md
```diff
@@ -1631,6 +1631,168 @@ ThrottleSubmitter().start()
...
+           })
+job_spec = jpsi.JobSpecification({
+              'executable': 'echo',
+              'arguments' : ['foo', 'bar', 'buz']
+              'directory' : '/tmp/',
+              'stdin'     : '/dev/null',
+              'stdout'    : 'work.out',
+              'stderr'    : 'work.err',
+              'resources' : res_spec,
+           })
```
hategan, line 1699: This method of initializing the objects is not in the API, so I am not sure what message we want to send here:
- that frameworks have quite a few liberties here at the expense of source-level compatibility
- or that this IS in fact what the API is saying should be how JPSI objects are initialized

SteVwonder (approved):
LGTM!

hategan (approved):

hategan:
OK, I don't want to hold this. When/if we decide on a better way to initialize the job info, we can update the examples as needed.

Issue #10, opened 2020-11-07
Consider adding further exceptions to submit() 


andre-merzky:
... in order to distinguish between EAGAIN types of errors and others.

hategan:
Question for @andre-merzky and @SteVwonder:

do we think:
a. `InvalidJobException` can have EAGAIN subtypes?
   - perhaps related: would a "too many jobs in the queue" error be an `InvalidJobException` or a `SubmitException`?
b. `SubmitException` can have EAGAIN subtypes?

andre-merzky:
> do we think:
> a. `InvalidJobException` can have EAGAIN subtypes?
> 
>     * perhaps related: would a "too many jobs in the queue" error be an `InvalidJobException` or a `SubmitException`?

I would argue that `EAGAIN` should be the decision criteria on `InvalidJobException` vs. `SubmitException`: an invalid job will remain invalid and trying again will not change that.  For example: if a job requests a node type which does not exist on the target machine, then that is invalid, but `too many jobs in the queue` can apply for a perfectly valid job, it just happens that the job cannot be submitted right now because of policy issues.

> 
> b. `SubmitException` can have EAGAIN subtypes?

Yes, I would expect so.  I can imagine cases where that job would fail again and again, for example if the queues are in maintenance or the user's allocation expired and the scheduler refuses the job because of this - but even then, the submission *would* succeed when the queues are up again or on allocation renewal (meaning, the actual job request is valid).

My $0.02, Andre.

hategan:
Ok, so if I understand @andre-merzky correctly, `InvalidJobException` is "fatal" and `SubmitException` should have a `isTransient()` method. Yes?

I'd be cool with that.

andre-merzky:
> Ok, so if I understand @andre-merzky correctly, `InvalidJobException` is "fatal" and `SubmitException` should have a `isTransient()` method. Yes?
> 
> I'd be cool with that.

Essentially yes, that would work for me.

SteVwonder:
Sounds good to me!

andre-merzky:
This is now well defined in the spec - closing.

PR #96, opened 2021-01-28
Update submit exceptions

hategan:
This adds a note to `JobExecutor.submit()` that `InvalidJobException` is fatal and also adds `SubmitException.isTransient()` as discussed in #10

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -391,16 +391,21 @@ status notifications about the job will be fired.
 ###### Exceptions:
 
 - `InvalidJobException`:
-    Thrown if the job specification cannot be understood. In principle,
-    the underlying implementation / LRM is the entity ultimately
-    responsible for interpreting a specification and reporting any errors
-    associated with it. However, in many cases, this reporting may come
-    after a significant delay. In the interest of failing fast, library
+    Thrown if the job specification cannot be understood. This exception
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
Works for me - thanks

specification.md
```diff
@@ -391,16 +393,21 @@ status notifications about the job will be fired.
...
-    responsible for interpreting a specification and reporting any errors
-    associated with it. However, in many cases, this reporting may come
-    after a significant delay. In the interest of failing fast, library
+    Thrown if the job specification cannot be understood. This exception
+	is fatal in that submitting another job with the exact same details
+	will also fail with an `InvalidJobException`. In principle, the
+	underlying implementation / LRM is the entity ultimately responsible
+	for interpreting a specification and reporting any errors associated
+	with it. However, in many cases, this reporting may come after a
+	significant delay. In the interest of failing fast, library
```
SteVwonder, line 402: Nit: looks like these lines have a few extra spaces in the front.  Was that intentional?  Not sure how much it matters for the github markdown renderer.

SteVwonder (commented):
LGTM.  One minor comment:

specification.md
hategan: It's part of an item in a list, so it should be indented.

PR #97, opened 2021-01-29
change `overrideEnvironment` -> `inheritEnvironment`

### LINKED ISSUE #93: Override Environment
The spec states:
```
If this flag is set to `false`, the job starts with an empty environment.
The only environment variables that will be accessible to the job are the
ones set using `setEnvironment()`. If this flag is set to `true`, which
is the default, the job will also have access to inherited environment
variables. The precise nature of the inherited environment is left to the
implementation. In principle, this functionality is meant to allow
computing resources to pass various information to executing
applications, such as the location of a scratch directory in `$SCRATCH`,
while still allowing clients to define unrelated environment variables
```

I cannot reconcile my concept of `environment` with this statement, specifically with `If this flag is set to 'true', which
is the default, the job will also have access to inherited environment variables.`.  What is that supposed to convey?  Either the environment is empty, or it is not -- how is an `access to inherited environment variables` expected to look like if exactly that environment is purged per this flag?

I see another issue: what about env settings which are required by, say, MPI - those cannot possibly be set by the user.  Would setting the `flag` to `True` disable all MPI tasks?

I would suggest two changes:

  - keep the literal meaning of the flag: the original env is purged, the specified env replaces (overrides) it.
  - make the default `False`: getting env settings correct is not trivial, and the default env settings on a cluster (as determined by admins, batch system and launchers) is usually a good starting point for any application.

SteVwonder:
As discussed in #93, the semantics of the `overrideEnvironment` flag are
OK, but the name doesn't immediately evoke the right semantics in many
minds.  Change the name to `inheritEnvironment` to hopefully evoke the
correct semantics in users' minds.

Closes #93

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -659,10 +659,10 @@ arguments to the list by invoking `setArguments()` with a mutable list,
 then invoking `getArguments().add()`.
 
 
-<a name="jobspec-setoverrideenvironment"></a>
+<a name="jobspec-setinheritenvironment"></a>
 ```java
-void setOverrideEnvironment(boolean clearEnvironment)
-boolean getOverrideEnvironment()
+void setInheritEnvironment(boolean inheritEnvironment)
+boolean getInheritEnvironment()
 ```
 
 If this flag is set to `false`, the job starts with an empty environment.


hategan (approved):
Thanks!

PR #98, opened 2021-02-02
"Implemented" what I believe is the conclusion of #73. Also renamed o…

hategan:
…ne instance of "terminal state" to "final state".

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -42,6 +42,8 @@
 			- [Methods](#methods)
 		- [SubmitException](#submitexception)
 			- [Methods](#methods)
+		- [UnreachableStateException](#unreachablestateexception)
+			- [Methods](#methods)
 		- [ResourceSpec](#resourcespec)
 			- [Methods](#methods)
 		- [ResourceSpec](#resourcespec)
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -551,37 +553,64 @@ from a `QUEUED` state to a `COMPLETED` state. However, implementations
...
+entered the `FAILED` state while `targetStates` consists of `COMPLETED`),
+this method throws an
+[`UnreachableStateException`](#unreachablestateexception).
+
+<div class="imp-note">
+Implementations are encouraged to throw the `UnreachableStateException`
+as soon as it can be determined that the `targetStates` are unreachable
+and not necessarily when the job reaches a final state. However, whether
+it is possible to make such a determination before a final state is
+reached depends on the precise state model and on the exact time when
```
andre-merzky, line 577: The state model is precise, in the sense that we define it in this document - so I would suggest to remove that part.

andre-merzky (commented):
Thanks @hategan  - one comment, otherwise looks good to me.

specification.md
hategan: Are you sure? I meant it as a future-proof thing, in case we will need to update the state model. Our current state model is very carefully engineered to make this unnecessary, but I worry that it may end up insufficient.

andre-merzky: If we change the state model we probably have to change text in several places.  As it is, I would avoid the impression that the state model is flexible or exchangeable, or that the semantics somehow depends on the underlying batch system state model...

andre-merzky (approved):
Thanks :-)

Issue #74, opened 2020-12-21
move `cancel` to job


andre-merzky:
I suggest to move `cancel` to the Job  class.  At the moment, `wait` is on the Job, and `cancel` is on the Executor - this is inconsistent, and I frequently forget which is where.

hategan:
That's all great, but it means the job must keep a reference to the executor, which I'd like to avoid.

In principle:
- executor: does things (submit, cancel, etc.) that are specific to implementation
- job: has things (e.g., spec, states)

waitFor() is in Job because it's not really as much a "do thing" as it is a fancy form of "state == something" and is not specific to implementations (e.g. PBS, SLURM, etc.)

andre-merzky:
> That's all great, but it means the job must keep a reference to the executor, which I'd like to avoid.

How is that different from wait?

andre-merzky:
That's all great, but as said, I think it makes the API inconsistent.  The goal was to have a nice API, not a simple implementation (as we clearly state in the document).

hategan:
> I think it makes the API inconsistent.

Ok, but can we try to ground the discussion in something else than personal views?

andre-merzky:
> > I think it makes the API inconsistent.
> 
> Ok, but can we try to ground the discussion in something else than personal views?

It is my personal view I guess, but then again it is also not really subjective.  Can we try try to ground the discussion in something else than implementation issues? :-)  Sorry for the cheap shot, it was just offering itself... (**_\*blush\*_**)

hategan:
> It is my personal view I guess, but then again it is also not really subjective.

I don't doubt that. But let's try to articulate the "not really" part of "subjective" (aka. the objective). I could try to guess, but I'm fairly sure the result would not be reliable.

andre-merzky:
> But let's try to articulate the "not really" part of "subjective" (aka. the objective).

I tried to do that, let me try to phrase it clearer / more explicit.  

We have an Executor class to represent an LRM (or other executors) which manage jobs, and a Job class to  represent those managed jobs as stateful entities.  `executor.submit` is used to relate Jobs to Executors, and from that point on we expect the job state to be managed by the executor.  After that submit, there are two more methods defined which relate to job state (apart from inspection of said state which is always possible): `wait` and `cancel`.  It is inconsistent that one of those methods is defined on the Executor, and the other on the Job.  I suggest to define both on the Job.  

Defining both on the Executor would also be consistent, but I _subjectively_ like the other option more.

[...]

PR #100, opened 2021-02-03
Added Job.cancel() method.

hategan:
This is meant to address #74.


andre-merzky (approved):
Thanks, this looks good to me and resolves #74

Issue #33, opened 2020-11-13
Add constructors when only default constructor is there


hategan:
JobSpecification does not have a constructor. That's because only the default constructor is mandated. But it's hard to tell the difference between that and having only private constructors. So add default constructors to the doc to be clear.

hategan:
Fixed by PR #102

PR #102, opened 2021-02-05
Added constructors or notes that there are no declared constructors.

hategan:
This is meant to address #33, but goes a bit beyond that in that every class has a constructors section that either defines constructors or mentions why it doesn't.


andre-merzky (approved):
LGTM, thanks.

SteVwonder (approved):
LGTM!

PR #104, opened 2021-02-10
Extras updates

hategan:
This is mostly an update to the css/js used to display things cutely.

The major update is support for language binding accordions.
The other somewhat major update is that all language related choices are now synced (and saved in a cookie). So when you select one language in one place, it's selected in all places. This is based on the assumption that whoever reads the document most likely has one language in mind.

### DIFF
--- a/extras.css
+++ b/extras.css
@@ -1,21 +1,25 @@
 .imp-note {
     padding-left: 8pt;
-    padding-top: 24pt;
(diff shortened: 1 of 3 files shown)


andre-merzky (approved):
That's neat!

SteVwonder:
Very cool!

Issue #13, opened 2020-11-07
 think more about env var expansion in arguments and other places. 


andre-merzky:
The important issue is how much of a burden this is on implementations if we mandate it.

jameshcorbett:
I don't see it as much of a burden, but I don't see much of a benefit either. What does it allow that a user couldn't do themselves, without too much trouble? For instance, in Python, users could pretty easily use `os.path.expandvars` to perform the expansion, or something like `jobspec.environ = "/my/app/" + os.pathsep + os.getenv("PATH")`. The easier it is to do, the less burden on the implementer, but the less burden on the user if it is left out, too.

I think environment variable expansion should not be supported at all. I know plenty of libraries and workflow managers that don't support environment variable expansion. Furthermore, it also introduces complications arising from ordering, like @andre-merzky  mentioned [here](https://github.com/ExaWorks/job-api-spec/pull/50#discussion_r541797982). If we let users do their own environment-variable expansion, they can figure out those issues to their own satisfaction. Second, there could be instances where users actually don't want environment variable expansion. In cases like these, I think it is better to let users figure it out for themselves.

hategan:
> For instance, in Python, users could pretty easily use `os.path.expandvars` to perform the expansion,

This expansion refers to expansion on the compute node side. In other words, the environment variables that are being expanded may not be defined on the client side or they may have different values than on the client side. The user could have a pre_launch or wrapper script for this purpose, but the question is whether installing such wrapper scripts is a reasonable thing to ask the user to do in order to, say, add some directory to $PATH or figure out where $SCRATCH is.

jameshcorbett:
Ah OK I didn't see that anywhere in the discussion---it is spread out over many issues and PRs. That is more reasonable. So the use-case is that you don't want to override the Job's environment, you just want to add one or two things to whatever the default environment is on the target machine.

In that case I agree with Andre when [he said](https://github.com/ExaWorks/job-api-spec/issues/54#issuecomment-731845685)

> > ```
> > * if a user needs to do PATH=/path/to/my/app:$PATH, should it be required that a pre- script be used?
> > ```
> 
> 
> Yes

You mentioned "figuring out where $SCRATCH is". But isn't that what `inheritEnvironment=True` is for? The environment-variable expansion seems like it is for something else---where you want to extend an environment variable that will exist on the target machine---not for just inheriting environment variables that will exist on the target machine.

hategan:
[...]

PR #106, opened 2021-02-12
pre/post exec/launch scripts

### LINKED ISSUE #54: Pre, post, scripts, expansion, etc.
During the meeting on 11/20/20, there was an extended discussion that was started by PR #50.

It started with the motivation behind performing environment variable expansion for values in the environment dictionary of the jobspec. The use case behind that was the need to support doing the equivalent of `PATH=/some/path:${PATH}`.

This led to another question: What if the user wants to do `VAR1=value1` then `VAR2=value2/${VAR1}`. That is, reference a variable defined in the environment *earlier*. The problem is that, since the jobspec environment is stored in a dictionary, there is no clear order that would allow an implementation to know what was set *earlier*. To support that scenario, we would need to change the dictionary to a list of name/value pairs. This is all fine, but it is getting uncomfortably close to "why don't we just allow the user to specify a bunch of shell commands to be executed before the executable is launched?". That is, why not just dump any fancy processing in the API and have pre/post-execs instead. That would allow a lot of flexibility.

But then, the law of slippery sloping would dictated that maybe we should have a rather simplified API that specifies 3 scripts:
1. a pre script, running on the rank 0 node, which can contain all necessary `module load` commands, environment manipulations, etc.
2. an in-scipt, running on all ranks, which can contain the executable and all arguments
3. a post-script, also running on the rank 0 node, which can contain all necessary cleanup and aggregation

This would make the following features of the API unnecessary:
- executable/arguments (the in-script is sufficient)
- environment variable expansion in the environment dictionary
- possibly the whole environment dictionary (the users could simply set things in the pre-script)
- possibly output redirection, which could be handled by the in-script and post-scripts
- the directory; a cd command in the pre-script or in-script may suffice

Anyway, lots to think about...

SteVwonder:
Add setters/getters for the pre/post exec/launch scripts along with the
semantics of these scripts.

Closes #54

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -795,6 +795,59 @@ Path? getStderrPath()
 
 Set/get the paths to the standard stream files.
 
+<a name="jobspec-setprelaunch"></a>
+<a name="jobspec-setpreexec"></a>
+<a name="jobspec-setpostexec"></a>
+<a name="jobspec-setpostlaunch"></a>
+```java
+void setPreLaunch(Path script)
+Path? getPreLaunch()
+void setPostLaunch(Path script)
+Path? getPostLaunch()
+```
+
+Set/get the paths to the pre/post launch scripts.
+
+The `PreLaunch` script is sourced by the main process of the job before it uses
+the parallel launcher to execute the `executable`.  This script is intended for
+setting up the environment via external systems like `lmod` and `virtualenv` as
+well as for performing actions that should only be done once at the start of a
+job (e.g., creating a results directory).
+
+<div class="note impl">
+
+"Sourcing" the script means to execute the commands contained in the `PreLaunch`
+script within the current environment.  This is equivalent to the `dot` command
+as defined by POSIX.2. Any environment variables set or changed by the
+`PreLaunch` script SHALL be made available to the launcher. The parallel
+launcher should be configured to forward the environment from the calling
+process to ensure that environment variables set by `PreLaunch` are propagated
+to the rest of the job.
+
+</div>
+
+<div class="note user">
+
+Some environment changes can cause parallel launchers to malfunction.  For
+example, loading a module that changes out the interpreter for python or lua. In
+this case, `PreLaunch` may cause parallel launcher failures and accomplishing
+these environmental changes would be better done with a wrapper script around
+the `executable`.
+
+</div>
+
+
+The `PostLaunch` script is sourced by the process of the job that sourced the
+`PreLaunch` script. An explicit synchronization SHALL occur so that `PostLaunch`
+is sourced only after the exit of the `executable` on every rank of the job.
+The `PostLaunch` script may fail to run if the job walltime expires or if a
+single rank process hangs.
+
+<div class="note user">
+
+The `PreLaunch` and `PostLaunch` scripts SHALL be POSIX-compliant shell scripts.
+
+</div>
 
 <a name="jobspec-setresources"></a>
 ```java
--- a/specification.md
+++ b/specification.md
@@ -826,8 +826,14 @@ to the rest of the job.
 
 <div class="note user">
 
+Since the `PreLaunch` script is sourced before the parallel launch, it is only
+sourced by a single process. This process may be running on the login node or
+on a batch node, depending on the system and J/PSI implementation. In the case
+where code needs to run on every node or rank of a parallel job, it is
+advisable to instead use a wrapper script around the `executable`.
+
 Some environment changes can cause parallel launchers to malfunction.  For
-example, loading a module that changes out the interpreter for python or lua. In
+example, loading a module that changes the interpreter for python or lua. In
 this case, `PreLaunch` may cause parallel launcher failures and accomplishing
 these environmental changes would be better done with a wrapper script around
 the `executable`.
--- a/specification.md
+++ b/specification.md
@@ -796,8 +796,6 @@ Path? getStderrPath()
 Set/get the paths to the standard stream files.
 
 <a name="jobspec-setprelaunch"></a>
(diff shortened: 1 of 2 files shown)


SteVwonder:
Side note: I think if we go with this PR, then we should re-evaluate #50 and #13.

specification.md
```diff
@@ -795,6 +795,44 @@ Path? getStderrPath()
...
+<a name="jobspec-setpreexec"></a>
+<a name="jobspec-setpostexec"></a>
+<a name="jobspec-setpostlaunch"></a>
+```java
+void setPreLaunch(Path script)
+Path? getPreLaunch()
+void setPreExec(Path script)
+Path? getPreExec()
+void setPostExec(Path script)
+Path? getPostExec()
```
hategan, line 808: I seem to vaguely recall a discussion in which we may have agreed that pre- and post- exec can be done by a wrapper script, which does not require API support.

hategan (approved):

specification.md
```diff
@@ -795,6 +795,44 @@ Path? getStderrPath()
...
+void setPostLaunch(Path script)
+Path? getPostLaunch()
+```
+
+Set/get the paths to the pre/post exec/launch scripts.
+
+The `PreLaunch` script is sourced on rank 0 of the job before the parallel
+launch of the `executable`.  This script is intended for setting up the
```
andre-merzky, line 816: I have a terminology clash here, likely because we use the terms a bit differently in RCT?  

With `launcher`, we usually mean something like `mpirun`, `aprun`, etc,.  The job thus has no determined rank *before* launch - ranks are only a concept applicable *after* launch, and only then the qualifier (`on rank 0`) would make sense.

I appreciate the difference between running a script on `rank 0` or on all ranks.  We struggled with naming here, too, but now call `pre_exec` what runs before all ranks, and `pre_rank` what runs before a specific rank activates (we don't separate out rank 0).  

To be clear: I don't want to change anything in the spec, but want to clarify that the `pre_launch` is actually running *after* the mpirun, or whatever, well, _launches_ the job?

specification.md
```diff
@@ -795,6 +795,44 @@ Path? getStderrPath()
...
+files onto local storage).
+
+The `PostExec` script is sourced by every rank of the job after the exit of the
+`executable`.  The `PostExec` script SHALL run if the `executable` fails but it
+may fail to run if the job is killed due to exceeding its walltime.
+
+The `PostLaunch` script is sourced by rank 0 of the job after the exit of the
+`executable` and after the completion of the `PostExec` script on every rank.
+The `PostLaunch` script may fail to run if the job walltime expires, including
+if any of the `PostExec` scripts hang.
```
andre-merzky, line 835: Does `PostLaunch` imply synchronization, i.e., do we wait for *all* ranks to complete, or only for rank 0?  I would suggest the former, to allow the script to act upon global output data, for example.

andre-merzky (commented):
some questions...

SteVwonder:
Ok.  I think I addressed feedback from the call today and the comments above.  Ready for a second round of reviews.

specification.md
andre-merzky: In that case we should remove the `pre-exec` signatures.  We should then also mention `JPSI_RANK` which we intended to export to the task which supports `pre_exec` scripts.  That might be a  different PR though as we have discussed some other env vars too IIRC and might want to tackle this in one go.

specification.md
```diff
@@ -773,6 +773,59 @@ Path? getStderrPath()
...
+```
+
+Set/get the paths to the pre/post launch scripts.
+
+The `PreLaunch` script is sourced by the main process of the job before it uses
+the parallel launcher to execute the `executable`.  This script is intended for
+setting up the environment via external systems like `lmod` and `virtualenv` as
+well as for performing actions that should only be done once at the start of a
+job (e.g., creating a results directory).
+
```
andre-merzky, line 794: Thanks, this reads good to me.

One problem we repeatedly faced with our users is the assumption that `pre_launch` scripts run concurrently on the compute nodes.  Well, they don't - at best they run concurrently on the batch node, at worst they run serial on the login node...  We might want to document this - maybe not here in the spec though, but rather in the implementation docs?

andre-merzky (approved):
Thanks Steven, this looks good to me.

specification.md
hategan: I think you are looking at an outdated quote and @SteVwonder has already removed those signatures (although perhaps not the anchors).

andre-merzky: You are right, only the anchors are left.

specification.md
```diff
@@ -773,6 +773,59 @@ Path? getStderrPath()
 
 Set/get the paths to the standard stream files.
 
+<a name="jobspec-setprelaunch"></a>
+<a name="jobspec-setpreexec"></a>
```
SteVwonder, line 777: Note to self: remove the outdated exec tags

specification.md
SteVwonder: What are your thoughts on adding to the user note below something like the following:

```
Since the `PreLaunch` script is sourced before the parallel launch, it is only sourced by a single process.  This process may be running on the login node or on a batch node, depending on the system and J/PSI implementation.
```

specification.md
SteVwonder: Just pushed a fixup commit with those anchors removed.

specification.md
SteVwonder: I just pushed up a fixup commit with the suggested change above (plus an extra sentence or two).  Let me know what you think.  Happy to modify or back it out based on feedback.

andre-merzky (approved):
Thanks, no  further requests from my side.

PR #107, opened 2021-02-12
Additional Python constructors for JobSpec and ResourceSpecV1

### LINKED ISSUE #75: Constructors for JobSpecification and ResourceSpecification classes
In the context of the Python prototype implementation we began to [discuss](https://github.com/ExaWorks/jpsi-python/pull/2/files#r546603797) constructors for the Specification classes - those do not yet define constructors.  This issues moves that discussion to the specification scope.

hategan:
And now something that actually uses all that new javascript stuff. Also fixes #75.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -698,9 +698,9 @@ In Python, the default constructor is replaced with the following constructor:
 ```python
 JobSpec(name: str = None, executable: str = None,
         arguments: List[str] = None, directory: Path = None,
-        inheritEnvironment: bool = True,
-        environment: Dict[str, str] = None, stdinPath: Path = None,
-        stdoutPath: Path = None, stderrPath: Path = None,
+        inherit_environment: bool = True,
+        environment: Dict[str, str] = None, stdin_path: Path = None,
+        stdout_path: Path = None, stderr_path: Path = None,
         resources: ResourceSpec = None,
         attributes: JobAttributes = None)
 ```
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -688,8 +688,30 @@ JobSpec()
 
 Constructs an empty `JobSpec`.
 
-#### Methods
+<div class="lang-bindings">
+
+__Python__:
+
+In Python, the following additional constructor is defined:
```
SteVwonder, line 695: The argument-less constructor and this constructor are mutually exclusive in Python, right?  Do we want to also put the empty JobSpec constructor in this lang-bindings class?  Or maybe tweak the wording:

```suggestion
In Python, the following constructor is defined in lieu of the argument-less constructor:
```

specification.md
```diff
@@ -1124,6 +1146,25 @@ ResourceSpecV1()
 
 Constructs an empty `ResourceSpecV1` object.
 
+<div class="lang-bindings">
+
+__Python__:
+
+In Python, the following additional constructor is defined:
```
SteVwonder, line 1153: Same comment as above

SteVwonder (commented):
Generally LGTM!  The language bindings JS is very slick!  One comment about the multiple constructors in python.

specification.md
hategan: Thanks for catching this. Yes, I think the wording should be updated. The all-keyword-arg constructor can be called as nullary, but the text should make it clear that that's not an "additional constructor".

SteVwonder (approved):
LGTM!

specification.md
```diff
@@ -688,8 +688,30 @@ JobSpec()
...
+__Python__:
+
+In Python, the default constructor is replaced with the following constructor:
+
+<a name="jobspec-_**"></a>
+```python
+JobSpec(name: str = None, executable: str = None,
+        arguments: List[str] = None, directory: Path = None,
+        inheritEnvironment: bool = True,
+        environment: Dict[str, str] = None, stdinPath: Path = None,
```
jameshcorbett, line 702: Why camelCase rather than underscores?

hategan: because... silliness. Will update.

andre-merzky (approved):
LGTM

PR #108, opened 2021-02-19
Add Gemfile and Makefile

SteVwonder:
Buildling html locally is now as simple as:
```
bundle install
bundle exec make
```


hategan:
I haven't tested this, but +1. I've been using some ad-hoc Jekyll stuff to build locally. This looks much better than that.

andre-merzky (approved):
Great - works for me!

SteVwonder:
> I've been using some ad-hoc Jekyll stuff to build locally.

Sadly the resulting page is pretty bare bones, but it does the job.  If you want to add a Jekyll template, we can easily add Jekyll to the Gemfile.

PR #109, opened 2021-02-19
Clarify jobspec versioning and Flux vs J/PSI jobspec

### LINKED ISSUE #78: Clarification on spec versioning
The specification explicitly references the Flux JobSpec version 1, and uses it as basis for the job specification:

> This describes the version 1 of the job specification, which represents a request to run exactly one program. This version is based heavily on Flux's Jobspec V1

However, the specification has a `ResourceSpecificationV1`, but not a `JobSpecificationV1`.  I miss a clarification on how the versioning mentioned in the text is related to the versioning (or lack of) in the `ResourceSpecifiction` and `JobSpecification` classes.

SteVwonder:
Removed most references to V1 in the J/PSI jobspec.  While doing so, I noticed that we do have the `version` key in the YAML, which was undefined in the appendix.  I added a section for that.  While we might not have any current plans to add new versions of the J/PSI jobspec, it seems prudent to leave the key in case we find a need to in the future. 

Also added a sentence to each top-level key's section to describe how it connects to the `JobSpec`, `JobAttributes`, and `ResourceSpec` classes.

Closes #78

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1347,9 +1347,10 @@ listed in detail in the sections below.
 
 #### Resources
 
-The value of the resources key SHALL be a strict list which MUST define
-either node or slot as the first and only resource. Each list element
-SHALL represent a **resource vertex** (described below).
+The `resources` key of the J/PSI jobspec contains the data from the
+`ResourceSpec` class.  The value of the resources key SHALL be a strict list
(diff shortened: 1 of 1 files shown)


hategan (approved):
+1

andre-merzky (approved):
LGTM, thanks

PR #112, opened 2021-02-25
Fixed what looks like some formatting that went quite wrong

hategan:


SteVwonder (approved):
LGTM!

Looks almost like emac's `fill-paragraph` command was run on this.  Good catch.

hategan:
> Looks almost like emac's `fill-paragraph` command was run on this. Good catch.

Or mc's (actually mcedit's) "Format paragraph" :)

PR #114, opened 2021-03-04
Add additional examples

### LINKED ISSUE #18: add examples
add examples of how one would use this API (and please, if you have any "how do you do x?", please add here)
  - [x] Submit X jobs and wait for them to complete in order of submission with the waitFor function 
  - [x] Submit X jobs but ensure only Y job are queued/running at a time (rolling window implemented with the jobStatusCallback) 
  - [ ] Submit a job, wait for it to be queued, then cancel it, then wait for it to complete/be killed.
  - [x] Submit a malformed or unsatisfiable job, then check for the error and print it out 
  - [x] Construct a job that uses all the various "knobs" of the resource and job specifications (with some verbose comments thrown in)

No longer applicable:
- [x] do the same but use the ListException from the submit call to do the same (this is related to bulk submission, which was removed)

### LINKED ISSUE #48: Clarify items in examples TODO
This was an in-line TODO item.

There's a PR that addresses some of the examples mentioned in that TODO item, but not all. The ones not addressed are not entirely clear. The reasons are listed below:

- [ ] do the same but use the ListException from the submit call to do
  the same

"the same" being ensuring that concurrent jobs are throttled to some maximum. I am assuming that whoever added this intended to track some "too many jobs in the queue" error from the LRM and act according to that. If so, it would be difficult to do this because we do not define a clear mechanism that could allow a clear way for a user to distinguish such an error from other errors. To state that differently, it's not impossible, but it isn't specified. One would have to rely on how the implementation happens to report errors of such type.

- [x] Submit a malformed or unsatisfiable job, then check for the error
  and print it out

Does anybody remember what the intent was here?

- [x] Construct a job that uses all the various "knobs" of the resource

I think this is rather generic. Can we be more specific?

SteVwonder:
#48 originally stated:
> I am assuming that whoever added this intended to track some "too many jobs in the queue" error from the LRM and act according to that. If so, it would be difficult to do this because we do not define a clear mechanism that could allow a clear way for a user to distinguish such an error from other errors. To state that differently, it's not impossible, but it isn't specified. One would have to rely on how the implementation happens to report errors of such type.

We don't have an explicit error for "too many jobs in the queue", but we do have a `isTransient` method for the `SubmitException`.  So I tweaked the example a bit to be a submission loop that continually attempts re-submission during transient issues using an exponential backoff algorithm.

Closes #18
Closes #48

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2077,6 +2077,42 @@ jex.submit(job)
 job.wait()
 ```
 
+#### Submit a job, check for transient error, retry if one occurred
+
+```python
+import time
+import math
+import random
+
+import jpsi
+
+def make_job():
+    job = jpsi.Job()
+    spec = jpsi.JobSpec()
+    spec.executable = '/bin/sleep'
+    spec.arguments = ['10']
+    job.specification = spec
+    return job
+
+def submit_with_exponential_backoff(jex, job):
+    times_attempted = 0
+    while(True):
+        try:
+            jex.submit(job)
+        except jpsi.SubmitException as se:
+            if not se.isTransient():
+                raise # re-raise to let caller see and handle it
+            times_attempted += 1
+            # https://en.wikipedia.org/wiki/Exponential_backoff
+            time.sleep(random.randint(0, math.pow(2, times_attempted) - 1))
+        else:
+            break
+
+jex = jpsi.JobExectorFactory.get_instance('slurm')
+job = make_job()
+submit_with_exponential_backoff(jex, job)
+job.wait()
+```
 
 
 ### Appendix D - Naming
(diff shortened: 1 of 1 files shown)


hategan (approved):
Looks good.

The following isn't really a comment about this PR, but I wonder how realistic the retry example would be in practice if one has to also consider the possibility of transient failure signaled through status change. And also whether we have the mechanism to accommodate that feature.

SteVwonder:
Thanks!

@andre-merzky : I just pushed a fix that changes `letter` -> `let`.  I didn't see any inline comment from you, but I assume that is the typo you were referring to.

PR #118, opened 2021-03-12
Add Path class definition

### LINKED ISSUE #105: `Path` is used in several function/method signatures but never defined
The type `Path` is used in several function/method signatures but never defined. 

### Suggested Clarification

I think most languages will already have a class type, so maybe it makes sense for us to just put a line somewhere in the spec that those should be preferred/used (like we did for the TimeInterval class).  If the language doesn't have a `Path` class/type, then maybe it can just fallback to being a string?

SteVwonder:
Closes #105

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -64,6 +64,7 @@
 			- [Constructors](#constructors)
 			- [Methods](#methods)
 		- [TimeUnit](#timeunit)
+        - [Path](#path)
 	- [Appendices](#appendices)
 		- [Appendix A - Job Specification Serialization Format](#appendix-a-job-specification-serialization-format)
 			- [Resources](#resources)
@@ -1401,6 +1402,13 @@ nearest integer value.
 Represents a time unit and must have at least the following units:
 `SECOND`, `MINUTE`, `HOUR`.
 
+### Path
+
+A class that allows users to specify a filesystem path in various formats.
+Implementations are encouraged to use standard library classes if available
+instead of re-implementing a custom `Path` class. If standard library classes
+are not available to represent filesystem paths, then the `String` class MAY
+be used instead.
 
 ## Appendices
 


hategan (approved):
Do we want to have the possibility of using, say, string in place of Path? If so, the constructor may be somewhat superfluous.

SteVwonder:
> Do we want to have the possibility of using, say, string in place of Path?

For Python, we can easily duck-type this to be a string or a path-like object.  For Java, do we just want to do overloading methods (one that takes `Path` and one that takes `String`)?

hategan:
> For Python, we can easily duck-type this to be a string or a path-like object. For Java, do we want to do overloaded methods (one that takes `Path` and one that takes `String`)?

My bad for not being clear. Should we allow for the possibility that, in some languages, a plain string will be used to represent a path? If so, that implies `Path == String`, which would make `Path::Path(String)` into `String::String(String)`, which is what looked superfluous to me.

SteVwonder:
Just pushed a commit that allows `Path` to be `String` if the implementation opts to not use the language's standard library definition (or the language doesn't have one).  @hategan : please let me know if that modification looks good to you or need more changes.

hategan:
> Just pushed a commit that allows `Path` to be `String` if the implementation opts to not use the language's standard library definition (or the language doesn't have one). @hategan : please let me know if that modification looks good to you or need more changes.

I think"SHALL" looks somewhat contradictory with respect to the previous paragraph in the sense that said paragraph gives the options as "Use std lib Path" or "implement Path", whereas the "SHALL" part strongly encourages a third option. So maybe "MAY" instead of "SHALL/SHOULD"?

SteVwonder:
Makes sense.  Just pushed a new commit with `SHALL` -> `MAY`.

andre-merzky (approved):
LGTM, Thanks.

Issue #115, opened 2021-03-04
Job ID usage


jameshcorbett:
The specification states that 

> The ID [of a job] is assigned automatically by the implementation when the Job object is constructed. The ID is guaranteed to be unique on the client machine. The ID does not have to match the ID of the underlying LRM job, but is used to identify Job instances as seen by a client application.

What is the use-case here? The ID "identifies Job instances as seen by a client application" but isn't this something users could do themselves, by associating an integer or a UUID with each Job? 

Also, the ID _can't_ feasibly match the ID of the underlying LRM job because the LRM ID isn't available until after the `Job` is constructed.

hategan:
> What is the use-case here? The ID "identifies Job instances as seen by a client application" but isn't this something users could do themselves, by associating an integer or a UUID with each Job?

Such an ID would not be available for any logging messages coming from within JPSI though. Which would make tracking jobs in logs pretty difficult if not impossible.

> Also, the ID _can't_ feasibly match the ID of the underlying LRM job because the LRM ID isn't available until after the `Job` is constructed.

Right. Hence the "does not have to match". It's written from the perspective of the user of the API rather than the implementer. But yes, it can't match the native ID given that it has to exist when the job is constructed.

jameshcorbett:
> Hence the "does not have to match".

I think something like "the ID is in no way connected with the ID of the underlying LRM job" would be more clear. Right now, the way it's phrased, it sounds like it might be up to the implementer to decide whether to match the `Job` ID to the LRM ID. But it's not.

> Such an ID would not be available for any logging messages coming from within JPSI though. Which would make tracking jobs in logs pretty difficult if not impossible.

If that's the use-case, though, the constraint could be reasonably relaxed so that the ID is only unique process-wide. And perhaps then relaxed further so that the ID is only guaranteed to be unique at any one particular point in time. But then in Python (for instance) why not just use `id()` and `repr()`, which is the standard thing to do?

But even aside from language-specific reasons, I think this ID stuff should be up to the implementation. If there isn't a well-defined and necessary reason for it, why keep it in there? It could always be added later if it was found to be necessary. Easier to add than remove.

hategan:
> [...] be up to the implementer to decide whether to match the `Job` ID to the LRM ID. But it's not.

That's a pretty strong statement that I'm not willing to make. For a traditional LRM, perhaps. In the most general case, I see no reason to exclude the possibility that a unique ID be pushed to a native ID.

[...]

PR #119, opened 2021-03-12
Specify logging behavior

jameshcorbett:
Based on some discussion on #115 I thought it might be good to specify how implementations should handle logging. 

On the one hand, users can do whatever job-state logging they want using the `addStatusCallback` methods of `Job` and `JobExecutor`. On the other hand, depending on the language it might be a little painful to actually make them do that. For instance, in Python, the library could set up a [logging.handlers.NullHandler](https://docs.python.org/3/howto/logging.html#library-config) and then let users configure the library's logging behavior through the normal mechanisms.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -23,6 +23,8 @@
   - [The Job API; Layer 0](#the-job-api-layer-0)
     - [Implementation Notes](#implementation-notes)
       - [Interaction with LRMs and Scalability](#interaction-with-lrms-and-scalability)
+    - [Behavior](#behavior)
+      - [Logging](#logging)
     - [JobExecutor](#jobexecutor)
       - [Methods](#methods)
           - [Exceptions:](#exceptions)
@@ -288,6 +290,15 @@ query interfaces (e.g.,  `qstat -a`) to get the status of all jobs and
 extract the information about the relevant jobs from the result.
 
 </div>
+
+
+### Behavior
+
+#### Logging
+
+Implementations must not perform any logging that cannot be configured and
+disabled by users. Users may perform their own job-state logging by using the
+`addJobStatusCallback` methods of `JobExecutor` and `Job`.
 
 
 ### JobExecutor


andre-merzky:
I am afraid I disagree with this PR.  Logging is a concern which is, to a first approximation, intern to the implementation.  I agree with the sentiment that a user should be able to disable logging, for both performance and privacy concerns.  But I certainly want to be able to log in an implementation, for example for debugging purposes, without having to make that logging facility user configurable.  Also, the statement as it stands still leaves the freedom of implementation on how logging could be configured, to what extent, etc - so it would likely not remove differences between implementations.

If logging is a concern (and I am not sure it is, given the state callback option you mention), then we should specify a limited set of well formatted log messages all implementations should issue.

jameshcorbett:
Good points, thanks. Would you be satisfied if "configurable" was removed from the PR? The main things I think are important are being able to disable logging, and being able to redirect the log messages to a different file or maybe some other kind of stream.

Like you, I also don't know how much logging is a concern. As far as I can tell, it depends on how many potentially useful log messages there are besides job state updates. If the callback could capture all of the interesting information (whatever that happens to be) then logging would probably not be all that helpful. Maybe this is something that we should just wait on, to see what different implementations would like to log, and then decide on it once we have more information? Unless other people can already see where logging would be useful/necessary for different implementations.

andre-merzky:
> As far as I can tell, it depends on how many potentially useful log messages there are besides job state updates.

I agree with that.  The problem is that `useful` is in the eye of the beholder.  When debugging, I usually log stuff like what cache gets discarded, what heartbeat message went missing, what triggered an increase in buffer allocation, when did the scheduler needed to do a full resource scan, and a gazillion other details which are just noise for the user: those messages relate to entities which are not exposed on API level and are not documented (in the user documentation).  Logging of state updates is really a minor part of those messages in most cases - from a debugging perspective, that is only interesting if unexpected state transitions occur.

Debugging is only one use of logs though (but the most important one for me personally).  


> Would you be satisfied if "configurable" was removed from the PR? The main things I think are important are being able to disable logging, and being able to redirect the log messages to a different file or maybe some other kind of stream.

That would result in :
```
Implementations must not perform any logging that cannot be disabled by users. 
Users may perform their own job-state logging by using the `addJobStatusCallback`
methods of `JobExecutor` and `Job`.
```

Yes, I would agree with that.  Being able to silence loggers is good practice.  Along the same lines, modules like j/psi should be, by default, silent on `stderr` and `stdout`.  Those things should be common practice, but it does not hurt to document them.

jameshcorbett:
>  Being able to silence loggers is good practice. Along the same lines, modules like j/psi should be, by default, silent on stderr and stdout. Those things should be common practice, but it does not hurt to document them.

I will change the commits to reflect that. Should there be something like an `enable_logging(boolean)` function or `JobExecutor` method?

andre-merzky:
> Should there be something like an `enable_logging(boolean)` function or `JobExecutor` method?

I would suggest to leave the mechanism for configuring (disabling) logs to the implementation.  Other details (where to log to, what log level to use, etc) are likely implementation specific also.  Specifically in Python it is customary to use Python's `logging` module which can globally set log levels via
```py
import logging
logging.disable(logging.INFO)
```

But other log systems may use config files or env variables.  Once your suggested text is in (user needs to be able to disable logger), any implementation is forced to document how the logger is disabled, which should be sufficient?

jameshcorbett:
> any implementation is forced to document how the logger is disabled, which should be sufficient?

I think that's good enough, so I wouldn't complain. But my thought was that if we force implementations to make logging optional, we might as well add a standard function/method for doing so, so that if users can guarantee that no matter what implementation they use, they can disable logs (or maybe enable them). But yeah, it would be tricky if it wasn't possible for a simple function call to do that. In Python I think it would be fine (make the function a wrapper around some calls to the `logging` module) but maybe in some implementations/languages it would be difficult.

> I also don't like to add config settings to the API

No, I totally agree. I didn't really have anything in mind when I said that users should be able to "configure" the logging of implementations, aside from stream redirection, which I think is fine to let implementations do themselves.

andre-merzky (approved):
Sorry for letting this linger.  That PR looks good to me (but it is still marked as 'draft').

hategan:
@jameshcorbett can you rebase (or otherwise fix the conflicts) so we can merge this, please?

jameshcorbett:
Closing this for now b/c stale and not important.

PR #122, opened 2021-03-24
expand on the `getContext()` description

andre-merzky:
Also fixes a couple of smaller issues.


specification.md
```diff
@@ -942,12 +946,16 @@ int? getExitCode()
...
 
-<a name="jobstatus-getmessage"></a>
+<a name="jobstatus-getcontext"></a>
 ```java
-String? getMessage()
+Map<String, Any>? getContext()
 ```
 
-Returns the message associated with this status, if any.
+Returns additional, backend specific meta data associated with this status, if
```
SteVwonder, line 954:
```suggestion
Returns additional, backend specific metadata associated with this status, if
```

specification.md
```diff
@@ -942,12 +946,16 @@ int? getExitCode()
...
+any.  Those metadata may include details on the state transition, backend
+native job IDs, or other, non-standardized pieces of information.  An
```
SteVwonder, line 956:
```suggestion
native job IDs, or other non-standardized pieces of information.  An
```

SteVwonder (approved):
Generally LGTM! Just a few nits below.

specification.md
```diff
@@ -147,7 +151,7 @@ C](#bulk-submission).
 ## Layers
 
 There are at least three major ways in which a job management API can be
-used:
+used
```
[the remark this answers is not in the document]
andre-merzky: Ack! :-)

hategan:
What's the difference in purpose between `getContext()` and the existing `getMetadata()`?

andre-merzky:
> What's the difference in purpose between `getContext()` and the existing `getMetadata()`?

Good question - and in a different PR I wondered the same about `getMessage()` vs. `getMetadata()`.  I would think we need only one of those.

hategan:
See https://github.com/ExaWorks/job-api-spec/commit/a8dd4e949f9eb46971e839753ed81454a03fde58

Their description is very similar. It even says "metadata" for "getContext()". So I think this is a duplicate and should be reverted. 

We can discuss "getMessage" separately, but, for reference, it is meant to convey error messages on failed states, so I do not think it should be removed.

andre-merzky:
> See [a8dd4e9](https://github.com/ExaWorks/job-api-spec/commit/a8dd4e949f9eb46971e839753ed81454a03fde58)
> 
> Their description is very similar. It even says "metadata" for "getContext()". So I think this is a duplicate and should be reverted.
> 

You are probably right, and I don't mind reverting.  Sorry for the confusion.

> We can discuss "getMessage" separately, but, for reference, it is meant to convey error messages on failed states, so I do not think it should be removed.

I opened a ticket at #124.

PR #123, opened 2021-03-24
add python tab for constructor example

### LINKED ISSUE #117: Interface Specification: Ensure python snippets are available for interfaces with major diff between languages
Only some of the constructors/methods/etc have their python defintions

Ensure that we have a python definition for every interface/constructor/method etc

andre-merzky:
closes #117

The PR also convert tabs to spaces.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -12,75 +12,75 @@
 <!-- TOC depthFrom:1 depthTo:6 withLinks:1 updateOnSave:1 orderedList:0 -->
 
 - [A Portable Submission Interface for Jobs (J/PSI)](#a-portable-submission-interface-for-jobs-jpsi)
-	- [STATUS: EARLY DRAFT](#status-early-draft)
-	- [Introduction](#introduction)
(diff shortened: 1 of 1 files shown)


SteVwonder (approved):
LGTM!

hategan (commented):
I think we should do a global tab to spaces conversion rather than piece-wise. I hope that, no matter which side of the tabs vs spaces we are, we can all agree that mixing them is the worst.

andre-merzky:
> I think we should do a global tab to spaces conversion rather than piece-wise. I hope that, no matter which side of the tabs vs spaces we are, we can all agree that mixing them is the worst.

Agree.  They were mixed, and that didn't work well with the part I was working on - thus did the global replacement, there should be no tabs left.

andre-merzky:
@hategan : can we progress with this one, or do you have concerns?

hategan (approved):
Sorry.

andre-merzky:
no problem, and thanks!

PR #128, opened 2021-04-01
Fix examples

hategan:
This fixes a number of issues:
- "JobExecutor" was misspelled in a number of places as "JobExector"
- some code issues
- use JobExecutor.get_instance instead of JobExecutorFactory.get_instance
- a bunch of Job.specification references which should be Job.spec

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1933,14 +1933,14 @@ complete.
 ```python
 import jpsi
 
-jex = jpsi.JobExectorFactory.get_instance('slurm')
+jex = jpsi.JobExector.get_instance('slurm')
 
 def make_job():
     job = jpsi.Job()
     spec = jpsi.JobSpec()
     spec.executable = '/bin/sleep'
     spec.arguments = ['10']
-    job.specification = spec
+    job.spec = spec
     return job
 
 jobs = []
@@ -1966,9 +1966,9 @@ import jpsi
 
 class ThrottledSubmitter:
     def __init__(self):
-        self.jex = jpsi.JobExecutorFactory.get_instance('torque', '>= 0.2')
+        self.jex = jpsi.JobExecutor.get_instance('torque', '>= 0.2')
         # keep track of completed jobs so that we can submit the rest
-        self.jex.set_status_callback(self.callback)
+        self.jex.set_job_status_callback(self.callback)
         self.count = 0
 
     def make_job(self):
@@ -1976,7 +1976,7 @@ class ThrottledSubmitter:
 
     def submit_next():
         if self.count < N:
-            jex.submit(self.jobs[self.count])
+            self.jex.submit(self.jobs[self.count])
             self.count += 1
 
     def start(self)
@@ -1985,7 +1985,7 @@ class ThrottledSubmitter:
 
         # submit initial M jobs
         while self.count < M:
-            submit_next()
+            self.submit_next()
 
     def callback(self, job, status):
         if status.final:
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -2004,7 +2004,7 @@ Use the exception types to distinguish between re-triable or not
 ```python
 import jpsi
 
-jex = jpsi.JobExector()
+jex = jpsi.JobExecutor()
```
SteVwonder, line 2007: Note directly related to your spelling fix, but as a side question: What is a call to this constructor with no arguments supposed to return?  It isn't specified in the doc.  I'm guessing just a `local` connector/executor?

SteVwonder (approved):
LGMT!

specification.md
hategan: I don't think that's correct in any of the proposed mechanisms of getting an executor/adapter/connector instance.

So this will need fixing.

andre-merzky: A specific implementation might pick up the connector settings out of band?  For example, 
```py
from flux import jpsi
jes = jpsi.JobExecutor()
```

could choose to connect to the Flux service identified by the `FLUX_URI` env variable, similar to the native backend:
```py
import flux
handle = flux.Flux()
```

which does the same (but also accepts an optional URL).

andre-merzky (approved):
LGTM, thanks

specification.md
hategan: > ```python
> from flux import jpsi
> ```

Fair enough. It could do that.  But the `jpsi` package use there is awfully confusing. Everywhere else in the examples it is used as the `jpsi` package where, say, `Job` resides.

andre-merzky: > Everywhere else in the examples it is used as the `jpsi` package where, say, `Job` resides.

Apologies @hategan , I can't parse the above.  `Job` and `JobExecutor` would reside in the same module for any jpsi implementation, wouldn't they?  What's the confusion?

hategan: > `Job` and `JobExecutor` would reside in the same module

We talked about this a bit in the past and I argued that, while we can't control it, I'd rather we had a model in which not every connector contributor would have to re-write all the core classes.

So not in the scenario in which flux contributes an executor to an existing library.

andre-merzky: Ah, that's what you refer to, got it.  Well, in that case the argument changes to
```py
import flux
jex = flux.jpsi.JobExecutor()
```
but otherwise stands I guess.

Just for the record (and off-topic here): there are better ways to allow implementation reuse than asking the user to import two different modules and needing to remember what module to use for what part of the API...  brrr...

hategan: > ```python
> jex = flux.jpsi.JobExecutor()
> ```

Yes, that's more clear. But yes, argument stands. I usually like to also name it `FluxJobExecutor` in such cases, to avoid just that type of clash, but that's secondary.

> there are better ways to allow implementation reuse than asking the user to import two different modules and needing to remember what module to use for what part of the API..

Oh, agreed. And I think `psi.j.JobExecutor.get_instance("local"/"flux"/"rp")` or `psi.j.JobExecutor("local"/"flux"/"rp")` both fit the bill.

andre-merzky: > Oh, agreed. And I think `psi.j.JobExecutor.get_instance("local"/"flux"/"rp")` or `psi.j.JobExecutor("local"/"flux"/"rp")` both fit the bill.

as is the former example above.  Happy to provide code to show that.

hategan: I'm confused, but let's talk during meeting.

PR #144, opened 2022-01-21
Wilke/update package name

wilke:
Updated python package name in document. Changed every occurrence of jpsi to psij.
Added N=1 to first example in Appendix C to make it a working example.


hategan:
Awesome. Thank you!

PR #145, opened 2022-01-28
The code examples with alternate languages must not be indented, othe…

hategan:
…rwise the whole block ends up in a code section.


andre-merzky (approved):
LGTM, thanks

PR #147, opened 2022-04-12
Added a `Job.executor` property.

hategan:
Since `Job.cancel()` requires the job to hold an executor instance,
this property is implied but not explicit. Since jobs are submitted
using `JobExecutor.submit()`, the job does not get to see an executor
instance unless the executor explicitly communicattes a reference to
itself as part of the `submit()` call. However, if the property is not
defined in the spec, an executor implementation cannot know what property
to set unless it looks at the specifics of the implementation of the `Job`
class. But that also prevents the executor implementation from being
portable to other implementations of the PSI/J core classes.

An alternative solution might be to submit an attach jobs using 
`Job.submit(JobExecutor ex)` and `Job.attach(JobExecutor ex)`, but seems a 
bit forced given that the implementation of, say `Job.submit(ex)` would 
essentially be `this.executor = ex; ex.submit(this)`.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -602,6 +602,20 @@ the above constructor, the job will be in the `NEW` state.
 
 #### Methods
 
+<a name="job-getexecutor"></a>
+```java
+void setExecutor(JobExecutor executor)
+JobExecutor? getExecutor()
+```
+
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
LGTM, can't see any reasons against it.

PR #151, opened 2023-03-15
Added folks known to have contributed to technical discussion and imp…

hategan:
…rovements of the spec.

Perhaps we should also have an acknowledgement section for reviewers/one-time contributors, etc.

PS: the above is based on my rather sketchy recollection, so please let me know if I missed somebody.


andre-merzky (approved):
Thanks for adding the names, and I like the alphabetic ordering.  We may want to mention on the Friday call that people are still invited to add their own (or other people's) names.

PR #154, opened 2023-03-21
Added some assumptions and goals.

### LINKED ISSUE #22: Add assumptions/goals/etc.
Things we could include in no particular order..
  - That we aim to make a minimal interface; advanced functionality is beyond the scope of this (version of the) API 
  - We would like the interface to be general and applicable to commonly deployed LRMs, cloud systems, etc. (I know some of this is said above) 
  - We are focused on executing a process (e.g., popen rather than function call)
  - That we intend for this interface to be used by various workflow systems and directly by applications 
  - That we base the API on lessons learned with SAGA, DRMAA, Globus, and others 
  - Do we want to set any goals about performance/scale? Presumably we want to aim to address exascale workloads and exascale machines (thousands of nodes) 
  - We consider allocation at the unit of a single job, no intention to dynamically update jobs

hategan:
It supersedes #44 and fixes #22.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -139,21 +139,45 @@ implementation.
 Specifically, the following aspects have informed the design in a significant
 fashion:
 
+- The interface is meant to be minimal and focused on submitting and managing
+jobs. Functionality that is outside that envelope, such as aggregating
+information about cluster queuing systems, simplifying the composition of jobs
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
Thanks for salvaging those snippets!

mturilli (approved):
LGTM. Thanks.

PR #155, opened 2023-03-21
TOC fixes. Anchors with the same name get a sequential index which th…

hategan:
…e previous TOC was not taking into account.


andre-merzky (approved):
Thanks!

mturilli (approved):
LGTM. Thanks.

PR #157, opened 2023-05-16
Small edits

hategan:
In design goals:
- emphasize a positive (that the primary use is for higher level tools) rather than a negative (that we don't prioritize user friendliness)
- similar for bulk calls: emphasize orthogonality and use the lack of bulk calls as an example


andre-merzky (approved):
Looks reasonable.

PR #158, opened 2023-05-16
Reformat to 80 cols

hategan:
Reformat to fit (snuggly) in 80 columns.

Where appropriate, also split method/property descriptions into a simple leading sentence followed by a detailed explanation.

Some minor edits, such as a missing period here and there.


andre-merzky (approved):
LGTM

PR #159, opened 2023-05-16
Removed the serialization appendix. There are a number of reasons for…

hategan:
… this:

- Not imposing specific choices when those choices are a matter of preference is an (unspoken - maybe we should speak it) design principle
- Should we be forced to impose such a choice, we should probably pay careful attention to picking something that will continue to be viable in the future (there is no good reason to believe that YAML will fare much better than XML).
- There are natural mappings from PSI/J classes to structured serialization formats. Such mappings associate property names in objects with homonymous keys in the serialized version, arrays/lists with lists, etc. The removed serialization format does not respect this. It introduces terms and concepts that would not be familiar to somebody who is familiar with the structure of PSI/J objects.
- The removed appendix occasionally mandates semantics that are not supported by the rest of the specification. Specifically, it is impossible to implement the resource requirements specified by this format with the current PSI/J specification.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -67,19 +67,12 @@
     - [TimeUnit](#timeunit)
     - [Path](#path)
   - [Appendices](#appendices)
-    - [Appendix A - Job Specification Serialization Format](#appendix-a---job-specification-serialization-format)
-      - [Resources](#resources)
-        - [Reserved Resource Types](#reserved-resource-types)
-        - [Resource Graph Restrictions](#resource-graph-restrictions)
-      - [Tasks](#tasks)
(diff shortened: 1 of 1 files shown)


hategan:
To be clear, this is a proposed change, not one that I am confident we have agreed on (although I believe there was some conversation that might have gone in that direction).

So please treat this not as a PR needing review, but as both a place to discuss this issue and possibly merge if agreed upon (let's call it a "prissue").

andre-merzky (approved):
Thanks for the spelled out rationale.  I can agree with that.

PR #160, opened 2023-05-16
Update layer2 section

hategan:


specification.md
```diff
@@ -241,10 +236,14 @@ should run.
...
-the rough functionality will be X, Y, Z
+**Note: the specification for the nested layer is not available at this time**
 
-- TBD
+- In addition to the remote layer, adds a mechanism to configure how pilot jobs
+are launched and managed. This may include the local layer executor to use for 
+launching the pilot jobs and a flexible mechanism that allows one to specify 
+various parameters for a pilot job management engine, such as limits on the
+number and sizes of pilot jobs, their scaling with respect to the user job 
+load, and circumstances under which pilot jobs are to be shut down.
```
andre-merzky, line 246: The text is fine as it is all phrased as 'we may do this or that', and I'll approve.  But for the record I'd like to comment that I see some of the listed items as out of scope for PSI/J as an *API*.  The wording seems to imply that a `pilot job management engine` and the responsibility for scaling pilots, would be part of what the Layer-2 API implements - but IMHO that would be beyond what an API should prove (which is pure abstraction and as little own semantics as possible).

To contrast: I would see PSI/J's role in Layer-2 to be as limited as providing a property or a callback on job startup which exposes a new PSI/J endpoint of that very job so that PSI/J can be used to push tasks into the pilot job.  The definition of pilot job number,  size, composition and lifetime would happen above PSI/J in that case.

Just my $0.02 :-)

andre-merzky (commented):
The PR removes the serialization appendix.  What's the rationale?  I am not to concerned about the serialization itself, but the text contained quite some details about the resource spec attributes and their semantics.  Should we reference the Flux docs instead maybe, or do you think that was redundant with other parts of the spec?

specification.md
andre-merzky: Thanks @hategan , appreciated

hategan: Thank you! Can you look at #159 when you have some time please?

hategan: @andre-merzky, I changed the Layer 2 paragraph to (hopefully) be more clear that pilot management is not mandated.

andre-merzky (approved):
Thank you @hategan , that reads well to me.

hategan:
Thank you @andre-merzky

PR #164, opened 2023-08-02
Added file staging

hategan:
This adds file staging as part of Layer 0.

File staging was initially part of Layer 1, but there is reasonable reason for existing Layer 0 executors to support file staging.
For example, the local executor is useful as a test executor and should support staging in order to allow testing of software without the complexity of remote execution. Another example is PBS, which natively supports staging directives in submit scripts.

There are some other changes, such as the removal of the "early draft" status and a move to something that is closer to a change log.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -742,8 +742,8 @@ Furthermore, implementations must, to the extent possible, simulate missing
 states. For example, if the implementation polls a LRM queue infrequently 
 enough such that the active state of a job is skipped between two polling 
 rounds, the job would appear to have jumped from a `QUEUED` state to a 
-`COMPLETED` state. However, implementations can, for example, introduce a 
-synthetic `ACTIVE` state change.
+`COMPLETED` state. In such cases implementations must introduce a synthetic
+`ACTIVE` state change.
 
 
 
--- a/specification.md
+++ b/specification.md
@@ -865,7 +865,8 @@ JobSpec(name: str = None, executable: str = None,
         attributes: JobAttributes = None)
         stage_in: Optional[Set[StageIn]] = None,
         stage_out: Optional[Set[StageOut]] = None,
-        cleanup: Optional[Set[Path]] = None)
+        cleanup: Optional[Set[Path]] = None,
+        cleanup_flags: StageOutFlags = StageOutFlags.ALWAYS)
 ```
 
 Creates an instance of `JobSpec` which allows properties to be initialized
@@ -1106,6 +1107,16 @@ a set of files that are to be deleted once the job completes and all relevant
 files are staged out but before the job is marked as `COMPLETED` (or `FAILED`).
 
 
+<a name="jobspec-set_cleanup_flags"></a>
+```java
+void setCleanupFlags(StageOutFlags flags)
+StageOutFlags getCleanupFlags()
+```
+
+Allows the controlling of the circumstances under which cleanup should occur. 
+By default, this flag should be set to `StageOutFlags.ALWAYS`. The value
+`StageOutFlags.IF_PRESENT` is ignored and the absence of files in the cleanup
+set when the cleanup step occurs must not result in an error.
 
 ### JobStatus
 
--- a/specification.md
+++ b/specification.md
@@ -216,6 +216,12 @@ filesystem.
 - Assumes that the client application is executed in an environment that has 
 direct access (i.e., does not require authentication) to a local LRM.
 
+- Enables file staging to and from job directory
+
+- Enables file cleanup
+
+- [Allows for standard output/error streaming if supported by the executor]
+
 
 ### Layer 1 (remote)
 
@@ -230,13 +236,6 @@ direct access (i.e., does not require authentication) to a local LRM.
     - Authentication and authorization
 
     - Encryption
-
-- Enables file staging
-
-- Allows for standard output/error streaming
-
-- Enables file cleanup (an alternative and perhaps more flexible way of doing
-this is to implement pre- and post-jobs in Layer 0)
 
 - May require user mapping if a system-wide service is deployed. User mapping 
 is a mapping of the authenticated user to a local user under which the job
--- a/specification.md
+++ b/specification.md
@@ -332,10 +332,6 @@ explicit runtime type-checks, but it is not required to do so.
 - Getters/setters can be replaced by properties, depending on what is customary 
 in the language in which the library is implemented.
 
-- Implementations may elect, at their discretion, to implement many of the 
-classes described in this document as read-only classes for which all 
-properties are set at construction time. 
-
 
 <div class="imp-note">
 
--- a/specification.md
+++ b/specification.md
@@ -13,7 +13,7 @@
 
 
 - [A Portable Submission Interface for Jobs (PSI/J)](#a-portable-submission-interface-for-jobs-psij)
-  - [STATUS: EARLY DRAFT](#status-early-draft)
+  - [Current Version: 0.2](#current-version-02)
   - [Introduction](#introduction)
     - [A Note about Code Samples](#a-note-about-code-samples)
   - [Motivation and Design Goals](#motivation-and-design-goals)
@@ -85,7 +85,16 @@
 
 <!-- /TOC -->
 
-## STATUS: EARLY DRAFT
+## Current Version: 0.2
+
+Version 0.1:
+- initial version
+- specification for Layer 0
+
+Version 0.2:
+- moves file staging from Layer 1 to Layer 0
+- adds file staging
+
 
 
 
--- a/specification.md
+++ b/specification.md
@@ -863,8 +863,9 @@ JobSpec(name: str = None, executable: str = None,
         stdout_path: Path = None, stderr_path: Path = None,
         resources: ResourceSpec = None,
         attributes: JobAttributes = None)
-        stageIn: Optional[StageInSet] = None,
-        stageOut: Optional[StageOutSet] = None,
+        stage_in: Optional[Set[StageIn]] = None,
+        stage_out: Optional[Set[StageOut]] = None,
+        cleanup: Optional[Set[Cleanup]] = None)
 ```
 
 Creates an instance of `JobSpec` which allows properties to be initialized
@@ -1203,8 +1204,8 @@ This class represents an enumeration and has no public constructors.
 ```java
 boolean isGreaterThan(JobState other)
 ```
-
 Defines a partial ordering on the states. 
+
 
 It is not possible to compare two final states—otherwise all state pairs are 
 comparable. Comparisons are transitive. The order is:
--- a/specification.md
+++ b/specification.md
@@ -61,11 +61,33 @@
     - [JobAttributes](#jobattributes)
       - [Constructors](#constructors-10)
       - [Methods](#methods-11)
-    - [TimeInterval](#timeinterval)
+    - [StageInSet](#stageinset)
       - [Constructors](#constructors-11)
       - [Methods](#methods-12)
+    - [StageOutSet](#stageoutset)
+      - [Constructors](#constructors-12)
+      - [Methods](#methods-13)
+    - [CleanupSet](#cleanupset)
+    - [Constructors](#constructors-13)
+      - [Methods](#methods-14)
+    - [StageIn](#stagein)
+      - [Constructors](#constructors-14)
+      - [Methods](#methods-15)
+    - [StagingMode](#stagingmode)
+      - [Constructors](#constructors-15)
+      - [Methods](#methods-16)
+    - [StageOut](#stageout)
+      - [Constructors](#constructors-16)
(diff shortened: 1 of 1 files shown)


ketancmaheshwari:
Is there a standard definition of staging behavior here or is that up for discussion? For example, would staging be a default or non-default behavior? Is it a copy or link operation or is it configurable? Can it be partial such as just of stage-out operations but not stage-in? Can it be partial among the set of files subject to staging such as based on size or may be a regex / wildcard?

hategan:
> Is there a standard definition of staging behavior here or is that up for discussion?

I would say that given the nature of the project, everything is up for discussion.

> For example, would staging be a default or non-default behavior?

I'm not sure I fully understand what you mean, but if a job does not specify files to stage, then nothing is staged.

> Is it a copy or link operation or is it configurable?

You can take a look at the changes, but yes, copy/link/move is configurable per file.

> Can it be partial such as just of stage-out operations but not stage-in?

A user can specify, for a job, files to stage out but no files to stage in.

> Can it be partial among the set of files subject to staging such as based on size or may be a regex / wildcard?

There is no complex filtering at this time (i.e., based on size or regex), but you can stage entire directories recursively. This is a good point though, and we should probably discuss if there is a need for such filtering that cannot be done otherwise. For example, stage-in filtering can be done by user code, since files to stage in are, by definition, accessible to the client side. Stage-out filtering may be a little more tricky, because user code doe not necessarily know what files the job will produce *before* the job runs. However, it is always possible to have the job run a post-processing step which filters relevant files into an output directory.

specification.md
```diff
@@ -301,6 +332,10 @@ explicit runtime type-checks, but it is not required to do so.
 - Getters/setters can be replaced by properties, depending on what is customary 
 in the language in which the library is implemented.
 
+- Implementations may elect, at their discretion, to implement many of the 
+classes described in this document as read-only classes for which all 
+properties are set at construction time. 
+
```
andre-merzky, line 338: It is an academic argument as long as we have only one implementation, but the statement above would limit application portability: an application using setters in one implementation could not easily switch to another implementation which has the proposed read-only semantics.

specification.md
```diff
@@ -567,15 +602,24 @@ that the following state model is adhered to.
 Job instances are, in this API, stateful objects.  A job's state can be
 inspected via the `job.getStatus()` method which will return a `JobStatus`
 instance on which the job's state is available as an attribute.  State
-transitions can also be received via callbacks
+transitions can also be received via callbacks. The transitions through 
+`STAGE_IN`, `STAGE_OUT`, and `CLEANUP` must be generated even if a particular 
+job does not specify any files for staging or clean up.
```
andre-merzky, line 607: Yes, I think this is important as it makes the transitions more predictable and the application code handling state changes becomes much simpler.

specification.md
```diff
@@ -567,15 +602,24 @@ that the following state model is adhered to.
...
 
 An implementation MUST ensure that job state transitions occur according to the
 following state model:  
   - A job is created in an initial state `NEW`.  
   - When the job is accepted by the backend for execution, it will enter the 
   state `QUEUED`.
+  - Before the job starts executing, it enters the
```
andre-merzky, line 614: I would suggest 'before job _execution_ starts' (and 'ends' two lines below) to distinguish this from the job instance life time.

specification.md
```diff
@@ -700,8 +746,9 @@ Furthermore, implementations must, to the extent possible, simulate missing
 states. For example, if the implementation polls a LRM queue infrequently 
 enough such that the active state of a job is skipped between two polling 
 rounds, the job would appear to have jumped from a `QUEUED` state to a 
-`COMPLETED` state. However, implementations can introduce a synthetic `ACTIVE` 
-state change.
+`COMPLETED` state. However, implementations can, for example, introduce a 
+synthetic `ACTIVE` state change.
+
```
andre-merzky, line 751: I would suggest to change `can` to `must` to be in sync with the requirement that `STAGE_IN` and `STAGE_OUT` transitions are generated even when no data are staged.

specification.md
```diff
@@ -1518,6 +1602,438 @@ attributes or not. It is, therefore, entirely possible for
...
+### StageInSet
+
+This class represents a set of stage-in directives, each represented by a 
+[`StageIn`](#stagein) object, that have no particular order. 
+
+<div class="imp-note">
+
+In languages with generic types and a `Set` type, a staging set could
+simply be `Set<StageIn>`, but implementations have the liberty to
```
andre-merzky, line 1614: I would suggest to change 'implementations have the liberty' to 'language bindings have the liberty' to ensure that the same API is implemented by different implementations in the same language.

andre-merzky (commented):
I know this is a draft, but left some comments anyway.  Thanks a bunch for working on this!

specification.md
hategan: Good point. And Python is rather unique in commonly having constructors with n >> 1 arguments due to kwargs. So perhaps it might be better to leave this out.

specification.md
hategan: Agreed. Thanks.

specification.md
hategan: I agree. My thought was that "must introduce a synthetic ACTIVE change" should not really apply if there is a real ACTIVE change. This should be reworded to clarify that *if the real one is missing*, a synthetic one must be introduced.

specification.md
hategan: Perhaps. I'm worried less about having to do `s/Set/StageInSet/` in select places than having to make algorithmic changes when switching from one implementation to another. Such a switch will most likely involve some changes. As long as the changes are simple and easily visible because an IDE underlines the wrong type, it is likely not a significant burden.

specification.md
hategan: Removed

specification.md
hategan: Done

specification.md
hategan: And done

specification.md
andre-merzky: I am more in the strict camp here.  Understand your point, and I assume in practice this is how it will go.  But it itches me that an implementor could come along and call her thing 'PSI/J compliant implementation' but I would need to figure out the subtle differences in the implementations.  Also, this seems to be the only point in the spec where we allow the implementor to choose a type - why here specifically?

andre-merzky: My comment should not block a merge - I will not veto based on this issue.

andre-merzky (approved):
Thanks for addressing my concerns!  I left a comment on the last open one, but please don't consider that blocking a merge!

specification.md
hategan: > But it itches me that an implementor could come along and call her thing 'PSI/J compliant implementation' but I would need to figure out the subtle differences in the implementations. 

I don't quite see having to change `StageInSet` to `Set<StageIn>` everywhere as a particularly problematic thing. It's well defined as opposed to semantic changes, which I think are more subtle. So I think it's more important for the semantics to be constrained rather than the various syntactical/morphological aspects.

> Also, this seems to be the only point in the spec where we allow the implementor to choose a type - why here specifically?

We don't really have collections in many other places. There are the spec argument list and env dict. But here we have two sets with distinct member types.

I think the point here is that in pre-generics Java, I would have added a new class. In post-generics Java, not so much, as `Set<>` is good and simple enough. I'm not as sure in Python, but I'd also lean towards `Set[]`. But then Python is special, and you could always say `StageInSet = Set[StageIn]`. So it seems to me that some degree of flexibility is necessary here.

andre-merzky: We may be talking about different things, sorry if I wasn't clear.  Or maybe we are, but then I am confused.  Let me try to rephrase please.

My concern is application portability.  If Java implementation 1 provides the method "`StageInSet()`" and java implementation 2 does not, then an application needs code changes to switch from impl 1 to impl 2.  Those changes might be trivial - but I want to argue that they should not be needed in the first place.  I would thus suggest that the existence of `StageInSet()` is defined by the *language binding*, not by the *implementation*.

I have no opinion whatsoever on the specifics of expressing stagein set construction in Java.

hategan: > We may be talking about different things, sorry if I wasn't clear. Or maybe we are, but then I am confused. Let me try to rephrase please.

Based on your answer, I think we are probably talking about the same thing.

> Those changes might be trivial - but I want to argue that they should not be needed in the first place. I would thus suggest that the existence of `StageInSet()` is defined by the _language binding_, not by the _implementation_.

I think my reasoning was that the changes are trivial and they are a lower cost than the combined cost of (possibly) overspecifying and needing language bindings for this. We can specify a binding for Java, but our bigger difficulty is specifying a binding for languages we don't know or that may not yet exist, since, with a Java language binding, we may solve this problem for Java, but for all other languages that might have a generic set, we will not.

So I do acknowledge the portability concern and it is indeed a concern. My view there is that switching PSI/J implementations would be a rare event in the lifecycle of a project AND that the cost would be low compared to the overall cost of making such a switch. In this particular case, I imagine that the change from `StageInSet` to `Set<StageIn>` would be a search/replace. More costly would probably be the cognitive load of determining that you need to do a search/replace for those types in the first place. However, on the latter point, I think such things would be inevitable, and any switch will involve some amount of testing and poking around.

hategan: I've updated the wording a bit to mandate the use of a generic set if it is available and encourage what is typical of the language used when implementing a custom class. I don't think it feels fully right, but I do think it better addresses your concern.

I would probably move ahead with this and let feedback come from actual implementations.

hategan: Thanks for spending the time to carefully think about this stuff and for a different perspective.

hategan:
An issue that came up is supporting file staging for batch schedulers.

Background: Some job schedulers support some basic form of staging. For example, PBS allows staging with the `-W stage[in|out]` attributes for `qsub` (see https://docs.adaptivecomputing.com/torque/2-5-12/Content/topics/commands/qsub.htm), Slurm can do staging via burst buffers (see https://slurm.schedmd.com/burst_buffer.html), and LSF uses '-stage' (see https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=options-stage). However, none of these reflect file staging states in an unambiguous way in the job states except Slurm having a STAGE_OUT state (see https://slurm.schedmd.com/squeue.html) but no STAGE_IN.

While the specification is kept separate from implementation issues, it should still allow for the implementation to actually be possible. So the issue at hand is if and how a potential batch scheduler executor could implement file staging.

There are a few possibilities:
1. Do not implement file staging for batch schedulers. Most supported file staging mechanisms in batch schedulers work at the job level (not the node level), so staging is most often equivalent to a copy to and from a persistent location at some point before or after the job runs. Consequently, the staging can be done from the job script or from outside the job entirely. The more interesting case of staging to and from compute nodes is not generally covered and may or may not be considered to be at the level at which PSI/J Layer 0 is supposed to work. The disadvantages of this is that there is a lack of parity on a significant issue between the batch scheduler executors and, e.g., the local executor, which may diminish the benefits of having the local executor as a stand-in testing executor (i.e., when using batch schedulers one may have to implement file staging outside of PSI/J, which would allow local/batch parity but would break batch/remote parity).
2. Implement file staging for batch schedulers either with native mechanism or with cp/mv/ln commands in the job script but acknowledge the (near) impossibility of correctly reflecting job states. That is because, in most cases, file staging would happen while the job is in an active scheduler state. Implementing out-of-band states is hindered by the following technical difficulties:
    * A potential network based solution of having a PSI/J implementation run a server to which jobs could connect to update status is made difficult by the fact that most login nodes have multiple network interfaces, only some of which are accessible from the compute nodes. There is no automatic, reliable, and efficient way of detecting which of the IP addresses associated with these interfaces is functional for the task at hand: probing them serially can run into minute-long timeouts while probing them in parallel can be resource intensive (maybe sending datagrams to all relevant IPs could work here?)
    * File based solutions are equally problematic in that tailing multiple files (one for each job) can be resource-intensive, whereas using a single file can run into atomicity/synchronization issues with writes on shared filesystems.
3. Implement file staging and finding a mechanism to correctly reflect job states. The downside here is that it would force a potentially difficult technical solution on the implementations.
4. Leave the decision to the implementation.

andre-merzky:
Late reply, sorry.

IMHO we should distinguish between (a) staging to the resource, (b) to the task sandbox (workdir), and (c) to node local storage.  (a) is only needed for layer 1 and 2 - but in that case is not enacted by the batch system, but need to happen *before* submission (I think).  (b) and (c) might be relevant to layer 0 - with the difference that (c) cannot possibly be enacted before job queuing as we do not know the exact set of nodes the job will run on (and/or don't have access to those nodes).  If the batch system does not support (c), the only chance I see is via the batch script - with all the caveats you point out.

I would suggest to specify (a) in layer 1/2 and (b) in layer 0, but not rely on the batch system, but rather on the PSIJ implementation to enact the staging.  An implementation MAY choose to support (c) as optimization.

Issue #129, opened 2021-04-01
Race condition in wait()


hategan:
There is an inherent race condition in `Job.wait()`, which is triggered by the "submit, wait for queued, cancel, wait for cancelled" example. The relevant code is:

```python
...
executor.submit(job)
...
job.wait([JobState.QUEUED])
...
```

If, at the time of the `wait` call, the job is in the `NEW` state, which the example expects, `wait` will return when the job reaches the  `QUEUED` state.
If, at the time of the `wait` call, the job is already `ACTIVE`, an `UnreachableStateException` will eventually be thrown.

Whether it's one of the other depends on a number of things, such as: 
 - if and how quickly the underlying executor moves jobs through a queue
 - what exactly is in-between the `submit()` and `wait()` calls.

For reference, with a local executor which sets the job to `ACTIVE` during `submit()`, the `UnreacheableStateException` will always be triggered.

### Suggested Solution

One could explicitly wait for all states after `QUEUED`, but that just seems tedious and contrary to the way we intended this to work, which is to not have such silly complexities.

andre-merzky:
I fully agree that the race exists - but that is not really avoidable in any system with concurrent threads of execution **when using a synchronous API** -- at least not with reasonable effort and performance.  I would argue that is is the responsibility of the user, and the proper solution is to use notifications:
```py
queued_signal = threading.Event()
def cb(job):
   if job.status.state == jpsi.QUEUED:
       queued_signal.set()
...
executor.submit(job)
...
queued_signal.wait()
...
```

hategan:
I agree with you.

And then, maybe I don't, in the sense that I think that if we can eliminate race conditions by careful API design, we could do so. 

So the problem here is not as much that a race condition exists, but that it is largely connector dependent. To me, that means a broken abstraction, which is contrary to the purpose of the API. If this occurred in equal measure for all connectors, I'd be more inclined to dismiss it as unavoidable.

In the spirit of seeing of this is avoidable, it strikes me that when I say `wait([JobState.QUEUED])`, I mean "don't return before the job is queued". In other words, I don't read that as waiting for a particular transition, that of `JobState.NEW -> JobState.QUEUED`, but more as a `ensure_that_the_job_has_gone_through(JobState.QUEUED)`. 

Point being, if we allowed wait(x) to return for any state >= x, we wouldn't quite have this problem.

andre-merzky:
[...]

PR #167, opened 2023-08-03
Fixed state ordering since QUEUED -> FAILED transitions are legitimate.

hategan:


specification.md
```diff
@@ -1134,8 +1134,8 @@ comparable. Comparisons are transitive. The order is:
   - `QUEUED    > NEW`
   - `ACTIVE    > QUEUED`
   - `COMPLETED > ACTIVE`
-  - `FAILED    > ACTIVE`
-  - `CANCELED  > ACTIVE`
+  - `FAILED    > QUEUED`
+  - `CANCELED  > QUEUED`
```
andre-merzky, line 1138: > "otherwise all state pairs are comparable."

It is now not obvious what the comparison `FAILED <> ACTIVE` would result in.
I would suggest:

```
  - `QUEUED    > NEW`
  
  - `ACTIVE    > QUEUED`
  - `COMPLETED > QUEUED`
  - `FAILED    > QUEUED`
  - `CANCELED  > QUEUED`
  
  - `COMPLETED > ACTIVE`
  - `FAILED    > ACTIVE`
  - `CANCELED  > ACTIVE`
```

andre-merzky: Actually, I take that back.  On a second look I don't really understand what the purpose of the change is.  Final states are supposed to be ordered after `ACTIVE` which is what the original text stated.  Since `ACTIVE > QUEUED` and the test states transitivity, `ACTIVE > QUEUED` is already implicit.  So the new text does not add anything, but removes `FINAL > ACTIVE`.  Why?

Or maybe I am reading this wrong?

hategan: `FAILED` and `ACTIVE` should not be comparable I think. That's because we can jump from QUEUED to FAILED without ever going trough ACTIVE. If two states are comparable, it means that we are certain that the smaller state must have come before the greater.

andre-merzky: I don't think I agree with that.  The purpose is to specify the `wait()` behavior.  So if we wait for `QUEUED` but find `FAILED`, then `wait` should return.  But the same holds if we wait for `ACTIVE` and find `FAILED`.

hategan: > I don't think I agree with that. The purpose is to specify the `wait()` behavior. So if we wait for `QUEUED` but find `FAILED`, then `wait` should return. But the same holds if we wait for `ACTIVE` and find `FAILED`.

Hmm. Not sure. If we wait for `ACTIVE` but the job goes from `QUEUED` to `FAILED`, I think that's an unreachable state situation, no? The contract of `wait()` returning being that it can guarantee that the job went through one of the target states.

hategan: Maybe we're complicating this too much and we should revert to having to explicitly enumerate the states we are waiting for.

Anyway, some context: #129

hategan: And I'll add something else. It seems to me that we might have two `FAILED` states, with quite different semantics. One is when the job runs and exits with a non-zero exit code. Some JAAPIs didn't even count this as failed (but completed). The other failed is when something bad happens and the job likely doesn't even get to run.

hategan: @andre-merzky and I had a live discussion and there were some ideas floated, which I'll attempt to list here:
- there is not much point for `wait(anything)` to wait if a final state has been reached
- `wait(ACTIVE)` should probably not raise `UnreachableStateException` if the job fails, since the order itself is insufficient to assert that `ACTIVE` didn't happen (at least in the current definition)
- we may want to get rid of `UnreachableStateException` and let users check the returned `JobStatus` for details of what caused `wait()` to return

Currently the official purpose of the order is that *"the system guarantees that no transition that would violate this ordering can occur"*. However, there is a bit of an ambiguity here. The partial ordering is a map `(s1, s2) -> T | F | null` with some properties (namely that `(s1, s2) -> T` implies `(s2, s1) -> F`, `(s1, s2) -> null` implies `(s2, s1) -> null`, etc.). We specify what cannot happen for two states for which `(s1, s2) -> F`, but we don't quite say what happens for `T` and `null` and how they differ. However, this can be figured out from the context to conclude that `(s1, s2) -> null` also means that the `s1 ---> s2` transition cannot happen. So basically the rules are:
- `(s1, s2) -> T`: allowed
- `(s1, s2) -> F`: not allowed
- `(s1, s2) -> null`: not allowed

The sentence *"the system guarantees that no transition that would violate this ordering can occur"* implies that anything else is allowed, which would include transitions for which `(s1, s2) -> null`. So that's not correct. I think it should read more along the lines of *"the system guarantees that a (possibly through some intermediate state(s)) transition `s1 -> s2` is only possible if and only if `s2 > s1`"*. That would force `FAILED > ACTIVE` because the current sentence doesn't.

andre-merzky: :-)

andre-merzky (approved):
This now reads correct and unambiguous to me, and also is a simply phrases as possible, IMHO. Nice!

PR #168, opened 2023-08-05
Move the `name` keyword arg from the first position in the Python

hategan:
constructor of `JobSpec` and have `executable` and `arguments` first so that we can write things like `JobSpec('ls', ['-al'])`


andre-merzky (approved):
Yeah, I am fine with this: common cases should be simple.

PR #169, opened 2023-08-05
Add and fix python constructors

### LINKED ISSUE #161: Consistency for Python constructors
For some classes with default constructors, such as `ResourceSpecV1`, language binding notes replace the default constructor with one where properties can be supplied as keyword arguments. However, some other classes, such as `JobAttributes` lack such a binding note.

The API should be consistent and generalize what is being done for `ResourceSpecV1` to all relevant classes.

hategan:
This adds a Python constructor for `JobAttributes`, in line with other classes, such as `JobSpec` and `ResourceSpecV1` (fixes #161).

It also fixes the type declarations for `JobSpec`.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -813,13 +813,17 @@ In Python, the default constructor is replaced with the following constructor:
 
 <a name="jobspec-_**"></a>
 ```python
-JobSpec(executable: str = None, arguments: List[str] = None, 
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
Works for me.

PR #170, opened 2023-08-05
Example fixes

### LINKED ISSUE #163: Fix "Run N Jobs while Throttling to M Concurrent Jobs" example
The code lacks proper synchronization (see https://github.com/ExaWorks/psij-python/pull/392).

### LINKED ISSUE #165: Examples incorrectly instantiate ResourceSpec, which is abstract
Some of the examples (e.g., "N Exclusive Nodes, Each with P Processes") contain lines like this:
```Python
res_spec = psij.ResourceSpec()
```

This does not work because `ResourceSpec` is abstract and cannot be instantiated. The examples should use `ResourceSpecV1` instead.

hategan:
The throttling example could occasionally lead to re-submission of the same job (because the counter increment wasn't protected by a critical section). This fixes it (fixes #163).

It also replaces some instantiations of `ResourceSpec` (which is abstract and cannot be instantiated) to `ResourceSpecV1` (this fixes #165).

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1943,7 +1943,7 @@ job.wait()
 ```python
 import psij
 
-res_spec = psij.ResourceSpec()
+res_spec = psij.ResourceSpecV1()
 res_spec.process_count     = 10
 res_spec.cores_per_process = 4
 res_spec.gpus_per_process  = 1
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
nice catch!

PR #171, opened 2023-08-05
Removed reference to native_id from `JobStatus.getMetadata`

### LINKED ISSUE #126: Remove mentions of the native id from JobStatus.getMetadata()

There's a mention of the native ID being passed back through state changes as part of the metadata in a JobStatus object. This was at a time when list()/attach()/ and the native_id property was not part of the spec. However, those have been merged (or are close to being so), so there is a persistent native ID property. making the native ID metadata attribute superfluous.

Remove mentions of the native ID from JobStatus.getMetadata()

hategan:
Fixes #126


andre-merzky (approved):
thanks!

PR #178, opened 2023-09-12
Changed `wait()` based on #167. This basically removes

hategan:
UnreachableStateException. Also restructured the text for `wait()` to hopefully make it easier to read/understand.


andre-merzky (approved):
I like this version - it does read clearer as we were hoping IMHO.

PR #180, opened 2023-11-17
Fix python constructor

hategan:
This adds launching related parameters to the python constructor. They appear to have been missing.

It also adds the set/getLauncher methods to `JobSpec`. Those were missing, too.


andre-merzky (approved):
Duh!

PR #182, opened 2024-03-19
Account not project

hategan:


hategan:
Note: this changes `projectName` to `account` in order to emphasize that it is for billing purposes, which it also clarifies.

It is on top of the file staging branch.

